PR #90, opened 2023-06-06
Fix unit tests

Isaac-DeFrain:
- fixes all unit tests
- adds a check to `IndexerState::add_block` to verify the block is not in the db before adding to the witness tree
- makes a few small modifications elsewhere to get tests to pass


jenr24G (approved):
Looks good! Nice Job!

PR #91, opened 2023-06-07
Fix genesis ledger parsing

trevorbernard:
---
  - https://app.asana.com/0/0/1204758919836130


Isaac-DeFrain (approved):
LGTM!

PR #92, opened 2023-06-07
Block log sorting

Isaac-DeFrain:
This PR adds a partial solution to our block ingestion woes. For a full solution, we will most likely need to implement pruning and re-rooting.

- implements sort on block height in `Parser`
- includes Jen's improved `get_precomputed_block`

---
  - https://app.asana.com/0/0/1204770208155640


jenr24G (approved):
Beautiful PR style, is this gonna be how we do PR's from now on?

LGTM I used your script version of this earlier for Trevor and it worked great!

Isaac-DeFrain:
> Beautiful PR style, is this gonna be how we do PR's from now on?

Thanks! I'm partial to it 😄
Updated with the no-renaming solution!

PR #95, opened 2023-06-09
Remove best chain field from state

Isaac-DeFrain:
Flamegraph exposed a lot of time being spent updating the `IndexerState::best_chain` field. Turns out, this was also causing a lot of unnecessary allocations too and the field is unnecessary!

This PR drastically improves performance

- removes the `best_chain` field from `IndexerState`
- replaces all uses of `best_chain` with a call to `root_branch.longest_chain()`
- modifies the server and tests accordingly

### DIFF
--- a/src/bin/mina-indexer-server.rs
+++ b/src/bin/mina-indexer-server.rs
@@ -11,7 +11,7 @@ use mina_indexer::{
     },
     state::ledger::{self, genesis::GenesisRoot, public_key::PublicKey, Ledger},
 };
-use tracing::{debug, error, event, info, instrument, Level};
+use tracing::{debug, error, info, instrument};
 use uuid::Uuid;
 
 #[derive(Parser, Debug)]
@@ -171,7 +171,7 @@ async fn main() -> Result<(), anyhow::Error> {
                 tokio::spawn(async move {
                     debug!("handling connection");
                     if let Err(e) = handle_conn(conn, block_store_readonly, best_chain, ledger).await {
-                        event!(Level::ERROR, "Error handling connection {}", e);
+                        error!("Error handling connection {e}");
                     }
                     debug!("removing readonly instance");
                     tokio::fs::remove_dir_all(&secondary_path).await.ok();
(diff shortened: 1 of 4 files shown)


src/bin/mina-indexer-server.rs
```diff
@@ -158,17 +155,22 @@ async fn main() -> Result<(), anyhow::Error> {
...
+
+                let mut leaves = indexer_state.root_branch.leaves.values().cloned();
+                let leaf = leaves.find(|leaf| &leaf.block.state_hash == best_chain.first().unwrap_or(&root_hash)).unwrap();
+                let ledger = leaf.get_ledger().clone();
+
                 tokio::spawn(async move {
                     debug!("handling connection");
-                    if let Err(e) = handle_conn(conn, block_store_readonly, best_chain).await {
+                    if let Err(e) = handle_conn(conn, block_store_readonly, best_chain, ledger).await {
                         event!(Level::ERROR, "Error handling connection {}", e);
```
trevorbernard, line 174: Can you make this call be `error!`

src/state/mod.rs
```diff
@@ -284,11 +278,17 @@ impl IndexerState {
...
+        if let Some(head_state_hash) = self.root_branch.longest_chain().first() {
+            // find the corresponding leaf ledger
+            for leaf in self.root_branch.leaves.values() {
+                if &leaf.block.state_hash == head_state_hash {
+                    res = Some(leaf.get_ledger());
+                    break;
+                }
+            }
         }
+        res
```
trevorbernard, line 291: Instead of break why not return the `Some(v)` there and `None` at the end. `let mut res` isn't necessary

trevorbernard (changes requested):

src/state/mod.rs
Isaac-DeFrain: Late night brain! Will correct

jenr24G (approved):
LGTM! Great catch with the performance improvements!

trevorbernard (approved):

PR #98, opened 2023-06-11
CLI flags defaults

jstavh:
Adding sane defaults for CLI flags.


jenr24G (changes requested):
I think you'll have better luck here: https://docs.rs/clap/latest/clap/ ! check out the `default_value_t` attribute!

PR #102, opened 2023-06-13
Sane defaults for CLI flags

jstavh:
Adding sane defaults for CLI flags.

### DIFF
--- a/src/bin/mina-indexer.rs
+++ b/src/bin/mina-indexer.rs
@@ -23,6 +23,11 @@ enum IndexerCommand {
 
 #[tokio::main]
 pub async fn main() -> Result<(), Box<dyn Error>> {
+    let indexer_folder = std::path::Path::new(concat!(env!("HOME"), "/mina-indexer"));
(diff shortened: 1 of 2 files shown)


jenr24G:
maybe we could put these directories in a `mina-indexer` folder in the user's home directory, and just name them as they are called in the `ServerArgs`. You would have to create the `mina-indexer` folder in the startup function if the defaults are used, though!

jstavh:
> maybe we could put these directories in a `mina-indexer` folder in the user's home directory, and just name them as they are called in the `ServerArgs`. You would have to create the `mina-indexer` folder in the startup function if the defaults are used, though!

@jenr24G I implemented your idea 😄

jstavh:
Wait, I need to add one more sane default to root_hash

jenr24G:
Last comment: i'd reccomend removing `default` from the names, also, make sure you run `cargo fmt` and `cargo clippy`!

jenr24G (approved):
LGTM!

PR #103, opened 2023-06-13
Simplify prune algorithm

Isaac-DeFrain:
- simplifies root branch pruning algorithm and ensures forks are dealt with appropriately
- adds `--prune_interval` cli flag
- adds `PartialOrd` and `Ord` impls for `Block`
- modifies tests to use sequential blocks over beautified
- removes the beautified sequential blocks

---
  - https://app.asana.com/0/0/1204820932108695


src/state/mod.rs
```diff
@@ -106,14 +101,10 @@ impl IndexerState {
...
         if let Some(k) = self.transition_frontier_length {
-            if let Some(interval) = self.prune_interval {
-                if self.blocks_processed % interval == 0 {
-                    info!("pruning transition frontier at k={}", k);
-                    self.root_branch.prune_transition_frontier(k);
-                }
-            } else {
-                info!("pruning transition frontier at k={}", k);
-                self.root_branch.prune_transition_frontier(k);
+            if self.root_branch.height() > 5 * k as usize {
```
jenr24G, line 104: can we make this constant configurable?

jenr24G (approved):
LGTM! I'd like to add a configuration option for how often we prune the root branch. Different use cases may warrant different pruning rates!

https://app.asana.com/0/1203669416123378/1204820393125452/f

trevorbernard (approved):
:shipit:

src/state/mod.rs
Isaac-DeFrain: I think that's a good idea!

Isaac-DeFrain: Will include with my fix for updating the best tip

trevorbernard (approved):
:shipit:

PR #104, opened 2023-06-13
Block ingestion binary

Isaac-DeFrain:
A lil app to ingest mainnet blocks and measure metrics


src/state/mod.rs
```diff
@@ -106,14 +101,10 @@ impl IndexerState {
...
         if let Some(k) = self.transition_frontier_length {
-            if let Some(interval) = self.prune_interval {
-                if self.blocks_processed % interval == 0 {
-                    info!("pruning transition frontier at k={}", k);
-                    self.root_branch.prune_transition_frontier(k);
-                }
-            } else {
-                info!("pruning transition frontier at k={}", k);
-                self.root_branch.prune_transition_frontier(k);
+            if self.root_branch.height() > 5 * k as usize {
```
jenr24G, line 104: same as with the other PR, i'd like to see a CLI option for this `5`

jenr24G (approved):
Looks good! I'll switch over to running this binary on mainnet, instead of Trevor's previous version!

PR #111, opened 2023-06-15
Restore state from db

Isaac-DeFrain:
This draft starts the process of restoring the indexer state from rocksdb. I've added some comments and TODOs where I intended to add code.

One really nice thing would be to integrate this with @jenr24's state serialization.

Some ideas I was hoping to implement _at a later time_ were multiple modes for the state restoration process. For example, there could be

- a "full verification" mode that starts at genesis and adds all blocks in order of `blockchain_length` (just like the state was created in the first place) (I'm really interested in how long this would take for all mainnet blocks since we bypass block parsing!)
- a "(verified) quick sync" mode that starts at a given (root) ledger and best tip and recreates the witness tree by getting the previous 290 or so ancestor blocks
- an "instant sync" that takes a best ledger and just deserializes the state and uses the db unchecked

Happy to discuss ideas and open to suggestions! Happy coding ya'll!


src/state/mod.rs
```diff
@@ -236,9 +237,117 @@ impl IndexerState {
...
+        let root_branch =
+            Branch::new_non_genesis(root_block.state_hash.clone(), root_block.blockchain_length);
+        let tip = Tip {
+            state_hash: root_block.state_hash.clone(),
+            node_id: root_branch.root.clone(),
+        };
+
+        Ok(Self {
+            mode: IndexerMode::Light,
+            phase: IndexerPhase::Watching,
```
Isaac-DeFrain, line 254: There’s an `InitializingFromDB` phase for this

src/state/mod.rs
```diff
@@ -236,9 +237,117 @@ impl IndexerState {
...
+            if let Some(length) = precomputed_block.blockchain_length {
+                if length > root_precomputed.blockchain_length.expect("already checked") {
+                    blocks_to_add.push(precomputed_block);
+                }
+            } else {
+                blocks_to_add.push(precomputed_block);
+            }
+        }
+        for block in blocks_to_add {
+            state.add_block(&block, false)?;
```
Isaac-DeFrain, line 347: Let’s bypass the witness tree for canonical blocks

Isaac-DeFrain (commented):
See comments

src/state/mod.rs
```diff
@@ -236,9 +237,117 @@ impl IndexerState {
...
+    #[instrument]
+    pub fn restore_from_db(db: IndexerStore) -> anyhow::Result<Self> {
+        // TODO
+        // find best tip block in db (according to Block::cmp)
+        // go back at least 290 blocks (make this block the root of the root tree)
+        // Q: How to compute ledger? Should we require it to do quick sync?
+        // iterate over the db blocks, starting at the root, adding them to the state with add_block(..., false)
+
+        // find the best tip
+        debug!("finding best tip in the provided database");
```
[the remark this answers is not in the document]
jenr24G: requiring this value will definitely reduce the compute time required to update the state, though it'll be a bit of effort to include it in the snapshot. I'd like to test the restoration without this value before making it required

src/state/mod.rs
jenr24G: thanks bestie got it

src/state/mod.rs
Isaac-DeFrain: Nice!

src/state/mod.rs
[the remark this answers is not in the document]
Isaac-DeFrain: If you think that's the best way forward. I'd imagine it would be an overall easier process to include this ASAP, but you are the one writing the code. Do it to it!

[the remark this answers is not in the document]
Isaac-DeFrain: Wouldn't we only need to include the state hash of the `best_tip`? Would that really be complicated to add to our snapshot?

PR #115, opened 2023-06-19
Filter canonical blocks on initial ingestion

Isaac-DeFrain:
This PR rethinks our initial ingestion of blocks. Instead of parsing and adding _all_ blocks in the startup dir to the witness tree, we instead _discover the longest canonical chain within these blocks_. Then we only parse and add _canonical_ blocks until we reach the highest canonical block in this directory, at which point we parse and add all successive blocks.

- implements the canonical chain discovery algorithm
- separates canonical and successive block paths in `BlockParser`
- adds `mode` and `phase` to `IndexerState`
  - `mode` dictates whether non-canonical blocks are stored in the db
  - `phase` allows us to change strategies depending on if we are initializing or watching
    - we may want to experiment with a longer `prune_interval` while initializing and a shorter one while watching
    - we may also want to experiment with aggregating ledger diffs and applying them in batches while initializing
- introduces `IndexerState::new_non_genesis()` to aid with our snapshotting efforts
- fixes some CLI flag defaults

---
  - https://app.asana.com/0/0/1204837873015977

### DIFF
--- a/src/bin/mainnet-test.rs
+++ b/src/bin/mainnet-test.rs
@@ -2,7 +2,7 @@ use bytesize::ByteSize;
 use clap::Parser;
 use mina_indexer::{
     block::{parser::BlockParser, BlockHash},
-    state::{ledger::genesis, IndexerState},
+    state::{ledger::genesis, IndexerMode, IndexerState},
     MAINNET_TRANSITION_FRONTIER_K,
 };
 use std::path::PathBuf;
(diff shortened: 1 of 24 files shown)


jenr24G (approved):
Absolutely fantastic, a true masterpiece right here. Thanks so much for all you've contributed to this project, and I hope to work with you again soon!

PR #116, opened 2023-06-20
Ingestion app watch cron job

Isaac-DeFrain:
Adds a "cron job" to download next blocks to the watch directory after initial ingestion


jenr24G (approved):
Great job my guy. absolutely powering ahead while you still can. Every PR I think is going to be your last ends up getting one upped by you yourself. Cool stuff, might wanna check the clippy annotations tho!

PR #120, opened 2023-06-20
Canonical ledger caching

Isaac-DeFrain:
This PR rethinks the way we calculate ledgers

- adds `canonical_tip` and `diffs_map` fields to `IndexerState` to enable us to cache ledgers and only apply diffs when needed
- adds `CanonicityStore` trait and corresponding column family to `IndexerStore` for persisting block statuses
- changes `root` and `best_tip` fields to `NodeId`s and adds block accessor functions

Unit tests are not included in this PR! They need to be added

---
  - https://app.asana.com/0/0/1204865169394584


Isaac-DeFrain:
I realized that clearing `diffs_map` was wrong. It has been corrected!

Isaac-DeFrain:
Investigating stack overflow! Things are heating up. Time to bust out flamegraph 🔥

Isaac-DeFrain:
As far as I'm concerned, this is G2G!
Adding mainnet ledger tests would be a good addition.
Also, we may want to experiment with different values for `CANONICAL_UPDATE_THRESHOLD` and `PRUNE_INTERVAL_DEFAULT` to find the right balance between ram usage and mainnet ingestion times. At their current values, I'm never seeing more than 2GB ram usage.

src/state/mod.rs
```diff
@@ -97,29 +127,40 @@ impl IndexerState {
...
             transition_frontier_length,
             prune_interval,
+            canonical_update_threshold,
             blocks_processed: 0,
             time: Instant::now(),
             date_time: OffsetDateTime::now_utc(),
         })
     }
 
+    #[allow(clippy::too_many_arguments)]
```
jenr24G, line 153: lol we'll need to fix this eventually

src/store.rs
```diff
@@ -114,6 +122,47 @@ impl LedgerStore for IndexerStore {
     }
 }
 
+impl CanonicityStore for IndexerStore {
```
jenr24G, line 125: can we add canonicity to the BlockStore with something like `updateCanonicity`?

Isaac-DeFrain: I'm all for that. Would you minding doing it? I really need to focus on CI stuff

jenr24G: got it bestie

jenr24G (approved):

PR #122, opened 2023-06-23
Short circuit canonical block diffs during bootstrap

Isaac-DeFrain:
This PR fixes the bootstrap ingestion performance problem
- instead of building the witness tree during bootstrapping, we simply calculate ledger diffs and apply them
- once we're done processing canonical blocks, we switch to building the witness tree


Isaac-DeFrain:
Still need to test ledger calculations.
I'd also like to have a ledger checkpoint config option.

jenr24G (approved):

PR #123, opened 2023-06-23
Adding slot field global_slot_since_genesis to PC block 

jstavh:
Adding field global_slot_since_genesis to precomputed block (corresponding struct and impl), which is consistent with mina-rs for global_slot_since_genesis. This adds support for staking ledgers (current and next) at epoch boundaries and therefore unblocking that issue. Also, some refactoring of tests and one new test.


Isaac-DeFrain (approved):
LGTM! I like the test too

PR #124, opened 2023-06-28
feat: index transactions

mxnnn:
## Changes

- Adds a `Key` interface
- Adds a `Transaction` interface to act as key variant
- Persists transactions when adding a block

### Notes
The Transaction storage key is structured as follows - `T-{Height}-{Signature}`
We use the signature as unique part of the key until we have a better way to identify transactions (e.g. hash).
The height is padded to 12 digits for sequential iteration.


trevorbernard (changes requested):
I see a couple of clippy warnings but other than that it looks good

trevorbernard (approved):

PR #126, opened 2023-06-28
Post Balances

jenr24G:
Changes our Ledger to use the Post Balances on each block to calculate account balances and nonces. This does not use the staged ledger diff Post Balance for the Coinbase Rewards, Transaction Fees, and Snark Worker Fees, which could be a bug in how MinaExplorer and the Archive Node calculate their balances, since this branch matches their balances exactly.

* create `UserCommand` type for the Post Balances
* create `Ledger::apply_post_balances`
* create `tests::state::ledger::apply_post_balances`

<hr>

Relevant Asana Tasks:

* https://app.asana.com/0/1203669416123378/1203770209987818/f
* https://app.asana.com/0/1203669416123378/1204886108856178/f


src/state/mod.rs
```diff
@@ -241,15 +242,136 @@ impl IndexerState {
...
+            if let Some(length) = precomputed_block.blockchain_length {
+                if length > root_precomputed.blockchain_length.expect("already checked") {
+                    blocks_to_add.push(precomputed_block);
+                }
+            } else {
+                blocks_to_add.push(precomputed_block);
+            }
+        }
+        for block in blocks_to_add {
+            // QUESTION: Prune blocks that create a dangling branch?
```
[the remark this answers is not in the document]
Isaac-DeFrain: I think the only way to do this in general is to check each block to see if it's parent exists in the db. If not, that's a problem and we can disregard it. If so, we could backtrack and count the length or break once we find an ancestor with a length or without a parent. Either way, I think it's splitting hairs...

src/store.rs
```diff
@@ -84,6 +84,10 @@ impl IndexerStore {
         &self.db_path
     }
 
+    pub fn iterator(&self) -> DBIteratorWithThreadMode<DBWithThreadMode<MultiThreaded>> {
```
Isaac-DeFrain, line 87: I'm questioning if this is necessary

Isaac-DeFrain (commented):
See comments

mxnnn (dismissed):

src/state/mod.rs
```diff
@@ -241,15 +242,136 @@ impl IndexerState {
...
+        global_slot_since_genesis: u32,
+    ) -> anyhow::Result<Self> {
+        // TODO
+        // find best tip block in db (according to Block::cmp)
+        // go back at least 290 blocks (make this block the root of the root tree)
+        // Q: How to compute ledger? Should we require it to do quick sync?
+        // iterate over the db blocks, starting at the root, adding them to the state with add_block(..., false)
+
+        // find the best tip
+        debug!("finding best tip in the provided database");
```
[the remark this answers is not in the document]
jenr24G: I agree with this now. It's pointless to write the algorithm to find these values when we can just include them in the snapshot

src/state/mod.rs
```diff
@@ -507,10 +632,11 @@ impl IndexerState {
     pub fn add_block(
         &mut self,
         precomputed_block: &PrecomputedBlock,
+        check_if_block_in_db: bool,
```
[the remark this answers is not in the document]
Isaac-DeFrain: I think we need this (or something else to achieve the same goal) for how we are doing the restoration from db. We need a way to distinguish between when we're adding blocks to the db and not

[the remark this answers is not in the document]
Isaac-DeFrain: But it can certainly be taken out of this "post balances" PR!

Isaac-DeFrain:
Can we get rid of the `Error::default()` thing in `src/state/ledger/mod.rs#L122`? Clippy is constantly complaining about it

src/state/mod.rs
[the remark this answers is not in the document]
jenr24G: let's move this to an issue @Isaac-DeFrain

jenr24G:
> Can we get rid of the `Error::default()` thing in `src/state/ledger/mod.rs#L122`? Clippy is constantly complaining about it

Jason fixed this!

trevorbernard (approved):

trevorbernard (changes requested):
There are some merge conflicts

src/state/ledger/mod.rs
```diff
@@ -51,6 +57,58 @@ impl Ledger {
         }
     }
 
+    pub fn apply_delegation(&mut self, delegator: PublicKey, new_delegate: PublicKey) {
```
Isaac-DeFrain, line 60: Why even check the old delegate if the outcome is the same either way?

src/state/ledger/mod.rs
```diff
@@ -51,6 +57,58 @@ impl Ledger {
...
+                }
+            }
+
+            account.delegate = Some(new_delegate);
+        }
+    }
+
+    pub fn apply_balance_update(&mut self, balance_update: BalanceUpdate, nonce: Option<i32>) {
+        if let Some(account) = self.accounts.get_mut(&balance_update.public_key) {
+            let nonce = if let Some(nonce) = nonce {
```
Isaac-DeFrain, line 74: Why not just conditionally set the nonce?

src/state/ledger/user_commands.rs
```diff
@@ -0,0 +1,117 @@
...
+            .diff
+            .t
+            .0
+            .t
+            .t
+            .commands
+            .iter()
+            .map(|command| command.t.clone())
+            .flat_map(|command| {
+                let source_nonce = match &command.data.t.t {
```
Isaac-DeFrain, line 41: All these `.t`s are gross! Maybe we can isolate all these accessors nicely into their own functions?

src/state/ledger/user_commands.rs
```diff
@@ -0,0 +1,117 @@
...
+                    staged_ledger_diff::UserCommand::SignedCommand(signed_command) => {
+                        signed_command.t.t.payload.t.t.common.t.t.t.nonce.t.t
```
Isaac-DeFrain, line 43: `signed_command.payload().nonce()` would be so much nicer!

src/state/ledger/user_commands.rs
```diff
@@ -0,0 +1,117 @@
...
+                    }
+                };
+                if let TransactionStatus::Applied(_, balance_data_versioned) = command.status.t {
+                    let mut delegation = false;
+                    let (fee_payer, source, receiver) = match command.data.t.t {
+                        staged_ledger_diff::UserCommand::SignedCommand(signed_command) => {
+                            let fee_payer =
+                                signed_command.t.t.payload.t.t.common.t.t.t.fee_payer_pk;
```
Isaac-DeFrain, line 51: `signed_command.payload().fee_payer_pk()` would be so much nicer!

src/state/ledger/user_commands.rs
```diff
@@ -0,0 +1,117 @@
...
+                            match signed_command.t.t.payload.t.t.body.t.t {
```
Isaac-DeFrain, line 52: `signed_command.payload().body()`

src/state/ledger/user_commands.rs
```diff
@@ -0,0 +1,117 @@
...
+    pub source_nonce: i32,
+    pub command_type: UserCommandType,
+    pub fee_payer: BalanceUpdate,
+    pub source: BalanceUpdate,
+    pub receiver: BalanceUpdate,
+}
+
+impl UserCommand {
+    pub fn from_precomputed(precomputed_block: &PrecomputedBlock) -> Vec<Self> {
+        precomputed_block
```
Isaac-DeFrain, line 30: Maybe `precomputed_block.commands()`?

Isaac-DeFrain (commented):
Some minor comments. Otherwise, LGTM!

src/state/ledger/user_commands.rs
Isaac-DeFrain: I mostly took care of this

Isaac-DeFrain (approved):
LGTM! But @mxnnn and @trevorbernard need to approve since they requested changes

trevorbernard (approved):

PR #127, opened 2023-06-29
Tests and error handling

jstavh:
Tests and error handling! So far within the ledger directory. Refactor the mod.rs slightly with error handling. Also fixes the clippy warning.

### DIFF
--- a/src/state/ledger/diff/account.rs
+++ b/src/state/ledger/diff/account.rs
@@ -165,8 +165,8 @@ impl std::fmt::Debug for UpdateType {
 
 #[cfg(test)]
 mod tests {
-    use super::{AccountDiff, PaymentDiff, UpdateType, DelegationDiff};
-    use crate::state::ledger::{command::{Payment, Command, Delegation}};
+    use super::{AccountDiff, DelegationDiff, PaymentDiff, UpdateType};
+    use crate::state::ledger::command::{Command, Delegation, Payment};
     use crate::state::ledger::PublicKey;
 
     #[test]
@@ -201,7 +201,6 @@ mod tests {
         ];
 
         assert_eq!(AccountDiff::from_command(payment_command), expected_result);
-
     }
 
     #[test]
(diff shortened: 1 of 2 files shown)


Isaac-DeFrain (commented):
LGTM!
Can you include a test comparing the `AccountDiff` from a block against the post balance calculation? I suggest using the genesis ledger and block 2 (as long as it has payments in it). We don't need to check all balances, just one that interacts with a payment should do

jenr24G (changes requested):
Looks great! I agree with Isaac about using the post balances as a source of truth! Can you fix the clippy warning? Thanks!

jstavh:
@Isaac-DeFrain @jenr24G To keep our PR's small, let's merge this one since you both think it looks good and then make a separate PR for the test you mention and also one for the `from_block_fees` function?

src/state/ledger/diff/account.rs
```diff
@@ -162,3 +162,132 @@ impl std::fmt::Debug for UpdateType {
...
+
+        let account_diff =
+            AccountDiff::from_coinbase(coinbase_receiver_result.into(), supercharge_coinbase);
+
+        let expected_payment_diff = PaymentDiff {
+            public_key: coinbase_receiver.into(),
+            amount: 1440 * (1e9 as u64),
+            update_type: UpdateType::Deposit,
+        };
+        let expected_account_diff = AccountDiff::Payment(expected_payment_diff);
```
Isaac-DeFrain, line 247: Interesting. This could explain why our nonces were off.

src/state/ledger/mod.rs
```diff
@@ -92,37 +98,48 @@ impl Ledger {
...
+                            Account::from_delegation(
+                                account_before,
+                                delegation_diff.delegate.clone(),
+                            )
+                        }
+                    };
+
+                    self.accounts.insert(diff.public_key(), account_after);
+                }
+                None => {
```
Isaac-DeFrain, line 132: How does this work with the creation of a new account?

src/state/ledger/diff/account.rs
Isaac-DeFrain: We're processing coinbases as payments? Payments should increase the nonce, but I don't think coinbases should. Not sure. Just a thought.

Isaac-DeFrain (approved):

jenr24G (approved):

PR #128, opened 2023-06-30
feat: added gql server & transactions query

mxnnn:
## Changes

This PR implements a GQL server with minimal structural changes.

- Adds `actix` & `juniper`
- Restructures RDB usage to be shared between services 
- Adds a transactions query with the following filters:
```rust
pub struct TransactionQueryInput {
    pub from: Option<String>,
    pub to: Option<String>,
    pub memo: Option<String>,
    pub date_time_gte: Option<DateTime<Utc>>,
    pub date_time_lte: Option<DateTime<Utc>>,
}
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -15,6 +15,9 @@ test = true
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
+actix-web = "4.3.1"
+actix-web-lab = "0.19.1"
+actix-cors = "0.6.4"
 anyhow = {version = "1.0.69"}
 async_executors = { version = "0.6.0", features = ["tokio_tp"] }
 async-trait = "0.1.64"
@@ -28,6 +31,8 @@ serde_json = {version = "1.0.92", features = [ "raw_value" ] }
 clap = { version = "4.1.4", features = [ "derive" ] }
 thiserror = "1.0.38"
 glob = "0.3.1"
+juniper = "0.15.11"
+chrono = "0.4.26"
 mina-serialization-types = { path = "./mina-rs/protocol/serialization-types", version = "0.1.0" }
 mina-signer = { path = "./mina-rs/proof-systems/signer", version = "0.1.0" }
 rocksdb = "0.20.1"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -31,8 +31,8 @@ serde_json = {version = "1.0.92", features = [ "raw_value" ] }
 clap = { version = "4.1.4", features = [ "derive" ] }
 thiserror = "1.0.38"
 glob = "0.3.1"
-juniper = "0.15.11"
-chrono = "0.4.26"
+juniper = { version= "0.15.11", default-features = false, features = ["chrono", "schema-language", "url", "uuid"] }
+chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
 mina-serialization-types = { path = "./mina-rs/protocol/serialization-types", version = "0.1.0" }
 mina-signer = { path = "./mina-rs/proof-systems/signer", version = "0.1.0" }
 rocksdb = "0.20.1"
--- a/src/gql/root.rs
+++ b/src/gql/root.rs
@@ -1,0 +1,109 @@
+use std::sync::Arc;
+
+use juniper::EmptyMutation;
+use juniper::EmptySubscription;
+use juniper::RootNode;
+use mina_serialization_types::staged_ledger_diff::UserCommandWithStatusJson;
+use mina_serialization_types::v1::UserCommandWithStatusV1;
+
+use crate::gql::schema::Transaction;
+use crate::gql::schema::TransactionQueryInput;
+use crate::store::IndexerStore;
+use crate::store::TransactionKey;
+
+pub struct Context {
+    pub db: Arc<IndexerStore>,
+}
+
+impl Context {
+    pub fn new(db: Arc<IndexerStore>) -> Self {
+        Self { db }
+    }
+}
+
+impl juniper::Context for Context {}
+
+pub struct QueryRoot;
+
+#[juniper::graphql_object(Context = Context)]
+impl QueryRoot {
+    #[graphql(description = "Indexer version")]
+    fn version() -> &str {
+        "0.1.1"
+    }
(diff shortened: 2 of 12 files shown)


Cargo.toml
```diff
@@ -28,6 +31,8 @@ serde_json = {version = "1.0.92", features = [ "raw_value" ] }
 clap = { version = "4.1.4", features = [ "derive" ] }
 thiserror = "1.0.38"
 glob = "0.3.1"
+juniper = { version= "0.15.11", default-features = false, features = ["chrono", "schema-language", "url", "uuid"] }
+chrono = { version = "0.4.22", default-features = false, features = ["clock"] }
```
Isaac-DeFrain, line 35: Is it possible to use `time` here? We're already using it and the last I recall there was a CVE out on `chrono`

Isaac-DeFrain (commented):
Just the one comment about `chrono`. Otherwise, LGTM!
I'll let someone with more GQL expertise review and pull the trigger.

src/state/mod.rs
```diff
@@ -51,7 +51,7 @@ pub struct IndexerState {
     /// needed for the possibility of missing blocks
     pub dangling_branches: Vec<Branch>,
     /// Block database
-    pub indexer_store: Option<IndexerStore>,
+    pub indexer_store: Option<Arc<IndexerStore>>,
```
trevorbernard, line 54: Why would this be optional?

trevorbernard (changes requested):

trevorbernard:
I'll make another pass at the PR when I have stable access to the internet so I can download the blocks

src/state/mod.rs
mxnnn: It was optional originally, but it definitely shouldn't be.

However, the changes necessary would make this PR noticeably bigger.

Cargo.toml
mxnnn: Ah yes of course, can do. 👍

src/gql/root.rs
```diff
@@ -0,0 +1,109 @@
...
+                }
+            }
+
+            transactions.push(transaction);
+
+            // stop collecting when reaching limit
+            if transactions.len() >= limit_idx {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
mxnnn: Wondering about the same.

Yes, it's a bit weird, and definitely differs from other GQL libs standards, but from what I could gather from reference implementations, this seems to be the way to go about it.

Cargo.toml
mxnnn: Actually, although I do agree, I'd propose sticking with chrono for now. Juniper removed the first class integration with `time`, so that'd mean writing custom scalars. Not sure we want to invest effort into that right now.
When it comes to the CVE, we should be good by not including default features.

src/gql/root.rs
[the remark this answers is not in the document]
mxnnn: Yea, applying filters definitely isn't as streamlined as I'd hoped & it'll get bloated quick once we add more queries.
If there is, I haven't found it yet - but I sure hope to find one lol.

src/gql/schema/transaction.rs
```diff
@@ -0,0 +1,102 @@
...
+                    timestamp: timestamp_to_datetime(timestamp),
+                }
+            }
+        }
+    }
+}
+
+// JSON utility
+fn sanitize_json<T: serde::Serialize>(s: T) -> String {
+    serde_json::to_string(&s).unwrap().replace('\"', "")
```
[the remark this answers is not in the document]
mxnnn: Serde serializes to a json string - so we'll need to remove escape chars.
Technically not necessary, multiple ways to achieve essentially the same, but kept it simple.

e.g.:
```json
{
   "from": "\"B62qre3erTHfzQckNuibViWQGyyKwZseztqrjPZBv6SQF384Rg6ESAy\"",
   "to": "\"B62qjYanmV7y9njVeH5UHkz3GYBm7xKir1rAnoY4KsEYUGLMiU45FSM\"",
   "memo": "\"E4YM2vTHhWEg66xpj52JErHUBU4pZ1yageL4TVDDpTTSsv8mK6YaH\"",
   "height": 750,
   "timestamp": "2021-03-19T04:00:00+00:00"
}
```

Cargo.toml
jenr24G: After doing some digging, it looks like `juniper` simply doesn't support `time`. Since this is the case, I agree with Max that it makes more sense to switch to time, thought @mxnnn you would need to replace the usage elsewhere in the code base or create an asana task for this before we could merge this PR

src/gql/root.rs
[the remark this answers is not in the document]
jenr24G: perhaps we could create a `Filter` trait and abstract over the specific filtering functionality?

src/gql/schema/transaction.rs
```diff
@@ -0,0 +1,102 @@
+use chrono::DateTime;
+use chrono::NaiveDateTime;
+use chrono::Utc;
```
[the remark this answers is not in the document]
jenr24G: since `juniper` doesn't support `time`, we'll need to go through the entire codebase and switch `time` out for `chrono`

src/gql/schema/transaction.rs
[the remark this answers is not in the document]
jenr24G: I think properly implementing `Display` for the `mina-rs` models makes a lot of sense, and can be a good first step to splitting off the code that we want, i.e. in the same vein as the `codec` from your `mina-indexer-bench` @mxnnn

src/state/mod.rs
jenr24G: This `Option` is used for testing, so that we don't need to set up a database connection in order to test nominally pure functions, like `IndexerState::add_block`. Emily's (what's her GH handle?) `MockBlockStore` would solve this issue, although also refactoring the code to better separate effectful and pure code would as well

Cargo.toml
Isaac-DeFrain: `chrono > time`
I'd vote for merging this and removing `time` later. It's not widely used so it will be a small change to do the switcharoo

trevorbernard (approved):

PR #130, opened 2023-07-05
Testing post-balance calculation logic

jstavh:
Adding additional testing where I set the initial balance of the account explicitly before applying the diffs. This allows us to verify that the post-balance calculation logic is working as expected. By explicitly setting the initial balance and applying the diff with a specific amount, we can check if the resulting balance in the ledger matches our expectations. These tests ensure that the ledger takes into account the existing balance of the account when applying the diffs and correctly updates the balance based on the diff type (deposit or deduction). Whereas the previous tests focus on testing the application of account diffs, while the new tests extend the testing to include the verification of balance calculations based on the initial balance and the applied diff.

https://app.asana.com/0/1203669416123378/1204865169394590

### DIFF
--- a/src/state/ledger/mod.rs
+++ b/src/state/ledger/mod.rs
@@ -302,4 +302,65 @@ mod tests {
 
         assert_eq!(account_after.delegate, Some(delegate_key));
     }
+
+    #[test]
+    fn apply_diff_payment_with_post_balance() {
+        let public_key =
+            PublicKey::from_address("B62qre3erTHfzQckNuibViWQGyyKwZseztqrjPZBv6SQF384Rg6ESAy")
+                .expect("public key creation");
+        let mut account = Account::empty(public_key.clone());
+        account.balance = Amount(10); // Set the balance explicitly
+        let mut accounts = HashMap::new();
+        accounts.insert(public_key.clone(), account);
+        let mut ledger = Ledger { accounts };
+
+        let ledger_diff = LedgerDiff {
+            public_keys_seen: vec![],
+            account_diffs: vec![AccountDiff::Payment(PaymentDiff {
+                public_key: public_key.clone(),
+                amount: 1, // Pass the underlying value
+                update_type: UpdateType::Deposit,
+            })],
+        };
+
+        ledger
+            .apply_diff(&ledger_diff)
+            .expect("ledger diff application");
+
+        let account_after = ledger.accounts.get(&public_key).expect("account get");
+
+        assert_eq!(account_after.balance, Amount(11)); // Assert against the balance
+    }
+
+    #[test]
+    fn apply_diff_delegation_with_post_balance() {
+        let public_key =
+            PublicKey::from_address("B62qre3erTHfzQckNuibViWQGyyKwZseztqrjPZBv6SQF384Rg6ESAy")
+                .expect("public key creation");
+        let delegate_key =
+            PublicKey::from_address("B62qmMypEDCchUgPD6RU99gVKXJcY46urKdjbFmG5cYtaVpfKysXTz6")
+                .expect("delegate public key creation");
+        let mut account = Account::empty(public_key.clone());
+        account.balance = Amount(20); // Set the balance explicitly
+        let mut accounts = HashMap::new();
+        accounts.insert(public_key.clone(), account);
+        let mut ledger = Ledger { accounts };
+
+        let ledger_diff = LedgerDiff {
+            public_keys_seen: vec![],
+            account_diffs: vec![AccountDiff::Delegation(DelegationDiff {
+                delegator: public_key.clone(),
+                delegate: delegate_key.clone(),
+            })],
+        };
+
+        ledger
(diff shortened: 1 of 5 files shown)


src/state/ledger/mod.rs
```diff
@@ -302,4 +302,65 @@ mod tests {
...
+            })],
+        };
+
+        ledger
+            .apply_diff(&ledger_diff)
+            .expect("ledger diff application");
+
+        let account_after = ledger.accounts.get(&public_key).expect("account get");
+
+        assert_eq!(account_after.balance, Amount(11)); // Assert against the balance
```
Isaac-DeFrain, line 332: Can you implement an `add(&self, amt: uint64)` method on `Amount` in `src/state/ledger/account.rs` and explicitly write the `Amount(11)` value as `initial_balance.add(ledger_diff.account_diffs[0].amount)`?

src/state/ledger/mod.rs
```diff
@@ -302,4 +302,65 @@ mod tests {
...
+        };
+
+        ledger
+            .apply_diff(&ledger_diff)
+            .expect("ledger diff application");
+
+        let account_after = ledger.accounts.get(&public_key).expect("account get");
+
+        assert_eq!(account_after.delegate, Some(delegate_key));
+        assert_eq!(account_after.balance, Amount(20)); // Balance should remain unchanged
```
Isaac-DeFrain, line 364: Then explicitly use `expect_balance` here?

Isaac-DeFrain (changes requested):

src/state/ledger/mod.rs
```diff
@@ -311,7 +311,7 @@ mod tests {
         let mut account = Account::empty(public_key.clone());
         account.balance = Amount(10); // Set the balance explicitly
         let mut accounts = HashMap::new();
-        accounts.insert(public_key.clone(), account);
+        accounts.insert(public_key.clone(), account.clone());
```
Isaac-DeFrain, line 314: Above line 313, write `let account_before = account.clone()`. Then you can eliminate this clone on 314.

src/state/ledger/mod.rs
```diff
@@ -327,8 +327,10 @@ mod tests {
             .apply_diff(&ledger_diff)
             .expect("ledger diff application");
 
+        let account_before = account.clone();
```
Isaac-DeFrain, line 330: See above comment

src/state/ledger/mod.rs
```diff
@@ -327,8 +327,10 @@ mod tests {
...
         let account_after = ledger.accounts.get(&public_key).expect("account get");
 
+        assert_eq!(account_before.balance, Amount(10)); // with cloning
```
Isaac-DeFrain, line 333: Instead of having two asserts with hardcoded values, make it a single assert
```rust
assert_eq!(account_after, account_before.add(ledger_diff.account_diffs[0].amount);
```

(you will also need to implement the `add()` method on `Amount`)

src/state/ledger/mod.rs
```diff
@@ -358,8 +360,10 @@ mod tests {
             .apply_diff(&ledger_diff)
             .expect("ledger diff application");
 
+        let account_before = account.clone();
         let account_after = ledger.accounts.get(&public_key).expect("account get");
 
+        assert_eq!(account_before.balance, account_after.balance); // with cloning
```
Isaac-DeFrain, line 366: Yes! You can get rid of line 368 now

Isaac-DeFrain (changes requested):

src/state/ledger/mod.rs
jstavh: Did this

src/state/ledger/mod.rs
jstavh: This required additional code so I'm adding it and will push it here soon for you to see.

src/state/ledger/mod.rs
```diff
@@ -327,11 +340,16 @@ mod tests {
             .apply_diff(&ledger_diff)
             .expect("ledger diff application");
 
-        let account_before = account.clone();
         let account_after = ledger.accounts.get(&public_key).expect("account get");
 
-        assert_eq!(account_before.balance, Amount(10)); // with cloning
-        assert_eq!(account_after.balance, Amount(11)); // Assert against the balance
+        if let AccountDiff::Payment(payment_diff) = &ledger_diff.account_diffs[0] {
```
Isaac-DeFrain, line 345: Please delete the `if ... else ...`, it is definitely and `AccountDiff`

Isaac-DeFrain (changes requested):

src/state/ledger/mod.rs
jstavh: Since payment_diff is not a direct field of AccountDiff, we may need to ```if else or ```match the enum variant to access it.

Isaac-DeFrain (approved):

PR #131, opened 2023-07-05
Fix canonical chain discovery bug

Isaac-DeFrain:
- fixes the bug in the canonical chain discovery algorithm
- simplifies command accessor functions and makes corresponding changes throughout

### DIFF
--- a/src/block/parser.rs
+++ b/src/block/parser.rs
@@ -164,8 +164,12 @@ impl BlockParser {
                         let prev_length_idx = length_start_indices[curr_start_idx - 1];
 
                         for path in paths[prev_length_idx..curr_length_idx].iter() {
-                            if extract_parent_hash_from_path(curr_path).unwrap()
-                                == hash_from_path(path).unwrap()
+                            // if there's a gap, skip the current length, or
+                            // if we found a parent, check the next lower length
+                            if length_from_path(path).unwrap() + 1
+                                < length_from_path(curr_path).unwrap()
+                                || extract_parent_hash_from_path(curr_path).unwrap()
+                                    == hash_from_path(path).unwrap()
                             {
                                 curr_path = path;
                                 curr_length_idx = prev_length_idx;
@@ -176,12 +180,16 @@ impl BlockParser {
                     }
                 }
 
-                let successive_idx = length_start_indices[curr_start_idx + 1];
+                let successive_idx = if curr_start_idx + 1 < length_start_indices.len() {
+                    Some(length_start_indices[curr_start_idx + 1])
+                } else {
+                    None
+                };
 
                 // curr_path represents a canonical block
                 info!(
-                    "Found canonical tip {} in {:?}",
-                    curr_path.file_name().unwrap().to_str().unwrap(),
+                    "Found canonical tip with hash {} in {:?}",
+                    get_state_hash(curr_path.file_name().unwrap()).unwrap(),
                     time.elapsed()
                 );
 
@@ -234,11 +242,13 @@ impl BlockParser {
                 canonical_paths.reverse();
 
                 // add all blocks successive to the canonical chain
-                for path in paths[successive_idx..]
-                    .iter()
-                    .filter(|p| length_from_path(p).is_some())
-                {
-                    successive_paths.push(path.clone());
+                if let Some(successive_idx) = successive_idx {
+                    for path in paths[successive_idx..]
+                        .iter()
+                        .filter(|p| length_from_path(p).is_some())
+                    {
+                        successive_paths.push(path.clone());
+                    }
                 }
             }
 
@@ -427,14 +437,14 @@ mod tests {
             .map(OsString::from)
             .map(|x| get_blockchain_length(&x))
             .for_each(|x| {
-                println!("{:?}", x);
+                println!("{x:?}");
             });
         Vec::from(FILENAMES_INVALID)
             .into_iter()
             .map(OsString::from)
             .map(|x| get_blockchain_length(&x))
             .for_each(|x| {
-                println!("{:?}", x);
+                println!("{x:?}");
             });
     }
 
--- a/src/block/parser.rs
+++ b/src/block/parser.rs
@@ -11,7 +11,6 @@ use std::{
     io::{prelude::*, SeekFrom},
     path::{Path, PathBuf},
     time::Instant,
-    u32::MAX,
     vec::IntoIter,
 };
 use tokio::io::AsyncReadExt;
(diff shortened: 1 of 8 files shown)


src/block/parser.rs
```diff
@@ -176,12 +180,16 @@ impl BlockParser {
...
+                } else {
+                    None
+                };
 
                 // curr_path represents a canonical block
                 info!(
-                    "Found canonical tip {} in {:?}",
-                    curr_path.file_name().unwrap().to_str().unwrap(),
+                    "Found canonical tip with hash {} in {:?}",
+                    get_state_hash(curr_path.file_name().unwrap()).unwrap(),
```
jenr24G, line 192: Can we return a `Result` from this function and use `?`'s instead of repeatedly calling `unwrap` everywhere?

src/block/parser.rs
```diff
@@ -164,8 +164,12 @@ impl BlockParser {
...
 
                         for path in paths[prev_length_idx..curr_length_idx].iter() {
-                            if extract_parent_hash_from_path(curr_path).unwrap()
-                                == hash_from_path(path).unwrap()
+                            // if there's a gap, skip the current length, or
+                            // if we found a parent, check the next lower length
+                            if length_from_path(path).unwrap() + 1
+                                < length_from_path(curr_path).unwrap()
+                                || extract_parent_hash_from_path(curr_path).unwrap()
+                                    == hash_from_path(path).unwrap()
```
jenr24G, line 172: let's use `?` instead of `unwrap`

src/block/parser.rs
```diff
@@ -234,11 +242,13 @@ impl BlockParser {
                 canonical_paths.reverse();
 
                 // add all blocks successive to the canonical chain
-                for path in paths[successive_idx..]
-                    .iter()
-                    .filter(|p| length_from_path(p).is_some())
-                {
-                    successive_paths.push(path.clone());
+                if let Some(successive_idx) = successive_idx {
```
jenr24G, line 245: maybe too much of a nitpick but `successive_idx.map(|idx| {...})` would be more idiomatic here

src/block/parser.rs
```diff
@@ -427,14 +437,14 @@ mod tests {
             .map(OsString::from)
             .map(|x| get_blockchain_length(&x))
             .for_each(|x| {
-                println!("{:?}", x);
+                println!("{x:?}");
```
jenr24G, line 440: this syntax looks so much better!

src/block/precomputed.rs
```diff
@@ -62,7 +67,7 @@ impl PrecomputedBlock {
         })
     }
 
-    pub fn cmds(&self) -> Vec<UserCommandWithStatusV1> {
+    pub fn commands(&self) -> Vec<UserCommandWithStatusV1> {
```
jenr24G, line 70: VOWELS

src/block/precomputed.rs
```diff
@@ -87,58 +90,31 @@ impl PrecomputedBlock {
...
-            .clone()
-            .inner()
-            .0
-            .inner()
-            .inner()
-            .commands;
+        let commands = self.commands();
         commands.iter().for_each(|command| {
-            let signed_command = match command.clone().inner().data.inner().inner() {
+            let signed_command = match UserCommandWithStatus(command.clone()).data() {
```
jenr24G, line 100: `mina-rs` in the TRASH

src/block/precomputed.rs
```diff
@@ -159,4 +135,18 @@ impl PrecomputedBlock {
             .t
             .t
     }
+
+    pub fn timestamp(&self) -> u64 {
```
jenr24G, line 139: this is like nanananana BATMAN but innerinnerinnerinnerinner TIMESTAMP

src/gql/schema/transaction.rs
```diff
@@ -9,13 +9,12 @@ use mina_serialization_types::staged_ledger_diff::UserCommandJson;
 
 use crate::gql::root::Context;
 
+#[allow(non_snake_case)]
```
jenr24G, line 12: but why? rust standard is to use snake case, we might as well go along with it

src/state/ledger/command.rs
```diff
@@ -31,59 +33,106 @@ pub enum Command {
     Delegation(Delegation),
 }
 
+#[derive(Debug, PartialEq, Eq, Clone, Serialize, Deserialize)]
```
jenr24G, line 36: we can derive `Hash` here too, might as well

src/state/ledger/command.rs
```diff
@@ -31,59 +33,106 @@ pub enum Command {
...
         precomputed_block
-            .staged_ledger_diff
-            .diff
-            .clone()
-            .inner()
-            .0
-            .inner()
-            .inner()
-            .commands
+            .commands()
```
jenr24G, line 48: c l e a n

src/state/mod.rs
```diff
@@ -399,25 +399,14 @@ impl IndexerState {
...
-                        .inner()
-                        .blockchain_state
-                        .inner()
-                        .inner()
-                        .timestamp
-                        .inner()
-                        .inner();
-
-                    for cmd in precomputed_block.cmds() {
+                    let tmstmp = precomputed_block.timestamp();
```
jenr24G, line 402: C L E A N

src/state/mod.rs
```diff
@@ -399,25 +399,14 @@ impl IndexerState {
...
+
+                    for cmd in precomputed_block.commands() {
                         indexer_store.put_tx(height, tmstmp, cmd)?;
                     }
                 }
 
-                // TODO store ledger at specified cadence, e.g. at epoch boundaries
+                // TODO: store ledger at specified cadence, e.g. at epoch boundaries
```
jenr24G, line 409: litchrally me

jenr24G (changes requested):
Looks Great! Let's just remove those `unwrap`'s and she's good to go!

src/state/ledger/command.rs
Isaac-DeFrain: Let's do it when we need it. Nothing warrants it yet AFAICT

src/state/ledger/command.rs
Isaac-DeFrain: 🧼

src/state/mod.rs
Isaac-DeFrain: 🧼

src/state/mod.rs
Isaac-DeFrain: Just in case ya'll forgot! Jk this is just to make vs code highlight it

src/block/parser.rs
Isaac-DeFrain: Added helpers

jenr24G (approved):
🚀

PR #133, opened 2023-07-07
Fix length comparison bug

Isaac-DeFrain:
- fixes length comparison bug in block parser


Isaac-DeFrain:
I also made takes in Asana to refactor/simplify this algorithm

trevorbernard (approved):

PR #145, opened 2023-07-26
New RocksDB for DelegationTotals-GQL

jstavh:
With these changes, we have a separate RocksDB database specifically designed to store delegation totals. This approach provides a clean separation of concerns, ensuring efficient retrieval of delegation totals for specific public keys and epochs without interfering with other blockchain-related data in the existing IndexerStore. Work in progress.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1857,6 +1857,9 @@ name = "hex"
 version = "0.4.3"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7f24254aa9a54b5c858eaee2f5bccdb46aaf0e486a595ed5fd8f86ba55232a70"
+dependencies = [
+ "serde",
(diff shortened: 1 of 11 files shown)


jstavh:
The last commit (excluding the clippy commit) so "conversions between gql and rust" focuses on types in the delegation gql -- f64 type does not implement the Eq and Hash traits, which are required by the juniper crate for GraphQL objects and i64 doesn't implement the required traits for being used directly as a GraphQL scalar value in the Juniper schema, so we can create a custom scalar type for representing the total_delegated field, and then convert the value between i64 and f64 when needed. This commit also tries to loop through the accounts to make a default epoch 1 but needs slightly more work along with a filler for the placeholder code for path to delegation_totals_db. With those two fixed, then we can switch from WIP to review.

PR #148, opened 2023-08-01
Fix canonical chain discovery bug

Isaac-DeFrain:
Recently, a regression was discovered in the canonical chain discovery algorithm. This PR fixes the bug and adds more tests. Specifically,

- checks for the existence of the ledger file
- fixes the canonical chain discovery bug
- adds multiple tests for different scenarios for canonical chain discovery
- fixes the `blocks_processed` count in the state when adding blocks


Isaac-DeFrain:
I know 91 files changed looks crazy! Most files are curated blocks for the new tests. The only substantive changes happen in `src/block/parser.rs`, `src/server/mod.rs`, and `src/state/mod.rs`.

PR #157, opened 2023-08-07
refactor BlockReceiver to trait

jenr24G:
Refactors the `BlockReceiver` struct to be a trait, so that Iggy's `loop` and `new_only` modes can be implemented under it. Moves the old `BlockReceiver` to the `FilesystemReceiver` struct, and makes necessary changes in `src/server.rs` and `tests`

<hr>

Asana Tasks:
https://app.asana.com/0/1203669416123378/1205188527548531/f


src/receiver/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+
+pub mod filesystem;
+
+#[async_trait::async_trait]
+pub trait BlockReceiver {
+    type BlockSource;
+    type Error;
+    async fn load_source(&mut self, source: &Self::BlockSource) -> Result<(), Self::Error>;
+    async fn recv_block(&mut self) -> Option<Result<PrecomputedBlock, Self::Error>>;
+}
```
trevorbernard, line 11: I would provide a synchronous trait along with an async wrapper. I would like more control of the threads

jenr24G: I don't think this makes sense. I could make the `load_source` signature sync, but the `recv_block` method is necessarily async as it entails waiting on an async worker for a result to come back. having `recv_block` as async gives the caller and the implementor more control as you can decide when to `await` at your convenience, rather than having to set up complex `poll` machinery on the user side. In addition this way allows me to construct a function `await_block(receivers: Vec<Box<dyn BlockReceiver>>) -> Option<Result<Block>>` that can be used to compose any number of implementors of this trait using `tokio::select!`. A synchronous trait would make the implementation of this function *much* more complex and error prone. I think this is a situation where async makes a lot of sense, though if you do insist, I will make the change

trevorbernard (approved):

PR #158, opened 2023-08-07
Mainnet test logging and filtering

Isaac-DeFrain:
- fixes `std::time::Duration` display error
- adds logging and block length filtering to the `mainnet-test` binary

### DIFF
--- a/src/bin/mainnet-test.rs
+++ b/src/bin/mainnet-test.rs
@@ -12,6 +12,10 @@ use tokio::{
     process,
     time::{Duration, Instant},
 };
+use tracing::level_filters::LevelFilter;
+use tracing_subscriber::prelude::*;
+
+// Assumed to be run on a block directory with almost all mainnet blocks.
 
 #[derive(Parser, Debug)]
 #[command(author, version, about, long_about = None)]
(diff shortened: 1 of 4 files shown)


src/bin/mainnet-test.rs
```diff
@@ -279,7 +297,7 @@ async fn main() -> anyhow::Result<()> {
         thread::sleep(Duration::new(180, 0));
     }
 
-    if !args.persist_db {
+    if !persist_db {
```
jenr24G, line 300: gotta remove database dir at the beginning 'stead of the end cuz of `ctrl-c`

src/block/parser.rs
```diff
@@ -100,7 +108,21 @@ impl BlockParser {
                     paths.len(),
                     display_duration(time.elapsed()),
                 );
-                info!("Searching for canonical chain in startup blocks");
+
+                if let Some(blockchain_length) = length_filter {
```
jenr24G, line 112: why do we need this?

trevorbernard (approved):

src/block/parser.rs
Isaac-DeFrain: If I run `mainnet-test` on a block dir with 275k blocks, I don't want to have to wait for canonical chain discovery to then just apply 10k blocks

Isaac-DeFrain: So this effectively makes it so that you only consider the first `blockchain_length` blocks in the dir

src/bin/mainnet-test.rs
Isaac-DeFrain: Beginning of what? You can't do this before the `for` loop and you shouldn't do it before the watch loop

PR #161, opened 2023-08-07
Delete and refresh db before ingestion

Isaac-DeFrain:
Deletes an existing test db for `mainnet-test`


Isaac-DeFrain:
@jenr24 I think this is what you meant

trevorbernard (approved):

PR #163, opened 2023-08-08
DelegationTotals GQL and new DB

jstavh:
WIP on creating a second database for delegations totals that will be publicly facing via GraphQL stakes.


src/delegation_totals_store.rs
```diff
@@ -0,0 +1,74 @@
+// RocksDB for DelegationTotals
+use crate::gql::schema::delegations::{DelegationTotals, TotalDelegated};
+use rocksdb::{Options, WriteBatch, DB};
+
+pub fn create_delegation_totals_db(path: &str) -> Result<DB, rocksdb::Error> {
```
jenr24G, line 5: Why can't this be a Column Family on the existing `IndexerStore`?

jenr24G (changes requested):

PR #164, opened 2023-08-09
Add GraphQL staking ledger functionality

trevorbernard:


jenr24G:
What's going on here? Could you add a description of the changes you made @trevorbernard ?

jstavh (approved):
LGTM! We used some code from here: https://github.com/Granola-Team/mina-indexer/pull/163

jenr24G (changes requested):
I still do not understand why we cannot use the existing `IndexerStore` database. This will be a breaking change for the Snapshot feature, and I don't think it's ready to merge until any potential regressions have been adressed

trevorbernard:
The reason why I didn’t use the existing DB is because you can’t import staking ledgers when the indexer is running because it already has a reference to all the other column families.

jenr24G (approved):

PR #165, opened 2023-08-09
Google Cloud Block Receiver

jenr24G:
Implements the `BlockReceiver` trait targeting Google Cloud buckets of Precomputed Blocks. Leverages Isaac's code from Iggy!
* Implement `GoogleCloudBlockWorker` to asynchronously check a Google Cloud bucket for blocks at a predefined interval
* Implement `GoogleCloudBlockReceiver` to spawn the above worker and allow a user to receive blocks on demand
* Add a test to ensure proper functioning of the system

<hr>

Asana Tasks Referenced
* https://app.asana.com/0/1203669416123378/1205188527548531/f


src/receiver/google_cloud/worker.rs
```diff
@@ -0,0 +1,234 @@
...
+
+            if let Ok(command) = self.command_receiver.try_recv() {
+                match command {
+                    GoogleCloudBlockWorkerCommand::Shutdown => {
+                        trace!("shutting down GoogleCloudBlockWorker");
+                        if tokio::fs::metadata(&self.worker_data.temp_blocks_dir)
+                            .await
+                            .is_ok()
+                        {
+                            tokio::fs::remove_dir_all(&self.worker_data.temp_blocks_dir)
```
jenr24G, line 90: huh weird this doesn't work. Had to add it again in `GoogleCloudBlockReceiver::drop`

tests/receiver/google_cloud.rs
```diff
@@ -0,0 +1,32 @@
...
+        bucket,
+    )
+    .await
+    .expect("google cloud block receiver constructor succeeds");
+
+    block_receiver
+        .recv_block()
+        .await
+        .expect("block is received successfully");
+    drop(block_receiver);
```
Isaac-DeFrain, line 31: Do we always have to explicitly `drop` the block receiver now?

src/receiver/google_cloud/mod.rs
```diff
@@ -0,0 +1,162 @@
...
+            command_sender
+                .clone()
+                .blocking_send(GoogleCloudBlockWorkerCommand::Shutdown)
+                .expect("shutdown command sends correctly");
+
+            if let Some(join_handle) = worker_join_handle {
+                join_handle.await.expect("worker fininshes successfully");
+            }
+        });
+        std::fs::remove_dir_all(temp_block_dir).expect("block dir exists");
```
Isaac-DeFrain, line 146: This is okay because it's just a temporary directory after all

src/receiver/google_cloud/worker.rs
Isaac-DeFrain: What happens if we eliminate it here?

Isaac-DeFrain (approved):
@jenr24 this PR is incredible! Superb, clean code. Simply nothing better.

tests/receiver/google_cloud.rs
jenr24G: I don't think so, I was just finagling with it to try and figure out why the temp folder wasn't deleted, this line is vestigial

src/receiver/google_cloud/mod.rs
jenr24G: we can remove the one in the command handler just fine, it doesn't do anything

PR #178, opened 2023-08-18
Add gstuil dependency to Docker

jstavh:
Add gstuil dependency in the dockerfile and installed curl and related.


jstavh:
To verify the code does what we want, I successfully built and ran docker with `docker build -t mina-indexer:latest . docker run --rm mina-indexer --help` then entered the container and checked for gsutil ```docker run -it --rm --entrypoint /bin/bash mina-indexer:latest

root@8820a24d7914:/# gsutil -v


Updates are available for some Google Cloud CLI components.  To install them,
please run:
  $ gcloud components update

gsutil version: 5.16

trevorbernard (approved):

PR #179, opened 2023-08-22
Add support for mina explorer graphql

jstavh:
To enable compatibility with the Mina Explorer GraphQL API, we incorporate a GraphQL client library for Rust into our project. This process involves creating *.graphql files that serve two important functions: first, they generate Rust code that corresponds to the GraphQL schema; second, they define the necessary Rust types to ensure compatibility with the schema. (Build X's so far are because of clippy complaining about unrelated code.)


trevorbernard:
This is in the wrong project. It belongs in the OCV app

PR #193, opened 2023-09-20
Improve DX

jenr24G:


robinbb (approved):

.gitignore
```diff
@@ -1,6 +1,4 @@
 /target
-.idea/*
```
trevorbernard, line 2: You should leave this in here. I'm experimenting with idea RustRover

jenr24G: hey how's that going? It seems great to me except that you can't connect to a vm over ssh

trevorbernard: I quite like it. I have a long history with intellij since my background is Java

robinbb (approved):

robinbb:
@trevorbernard I made the final commit, so GitHub is not counting my review, so I cannot merge. Care to give a review?

trevorbernard (approved):

PR #218, opened 2023-10-06
Problem: Save snapshot not adequately tested

trevorbernard:
There are no tests to determine if this feature is working


Isaac-DeFrain (approved):
LGTM

Isaac-DeFrain:
Why is rustfmt failing?

robinbb (approved):
My approval is based mostly on @Isaac-DeFrain 's approval, given that the only changes since then are syntax corrections for Clippy and a rebase.

PR #228, opened 2023-10-08
Add Indexer shutdown command

trevorbernard:


test
```diff
@@ -79,27 +81,30 @@ idxr server config 2>&1 |
 # Indexer server cli subcommand works
 idxr server cli --help > /dev/null
 
+# Temporary hack - create startup-blocks direct if it doesn't exist
```
jenr24G, line 84: can we put this in `${TMPDIR}/startup`?

test
```diff
@@ -79,27 +81,30 @@ idxr server config 2>&1 |
...
+mkdir $HOME/.mina-indexer/startup-blocks || true
```
jenr24G, line 85: we shouldn't need this since the tests override this to $BLOCKS_DIR

test
```diff
@@ -79,27 +81,30 @@ idxr server config 2>&1 |
...
 server cli \
-    --startup-dir $TMP/blocks \
-    --watch-dir $TMP/blocks \
+    --startup-dir $BLOCKS_DIR \
+    --watch-dir $BLOCKS_DIR \
     --initial-ledger tests/ledgers/mainnet-genesis.json \
     --is-genesis-ledger \
     --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ &
-IDXR_PID=$(echo $!)
+IDXR_PID=$(pgrep mina-indexer)
```
jenr24G, line 104: regression?

trevorbernard: `echo $!` was grabbing the wrong PID because of the subshell. So I reverted my suggestion

jenr24G: I'm seeing the same thing

Isaac-DeFrain (dismissed):
LGTM

Isaac-DeFrain (approved):

Issue #225, opened 2023-10-08
Problem: the indexer server isn't correctly initialized if the `startup-blocks` directory doesn't exist


trevorbernard:
The indexer isn't correctly initialized if the `startup-blocks` doesn't exist. The domain socket server will never get started meaning it never can accept commands.

```
rm -r $HOME/.mina-indexer/startup-blocks
./target/release/mina-indexer server cli \
    --initial-ledger tests/ledgers/mainnet-genesis.json \
    --is-genesis-ledger \
    --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ

2023-10-08T17:33:05.373775Z  INFO actix_server::builder: starting 10 workers
2023-10-08T17:33:05.374106Z  INFO actix_server::server: Tokio runtime found; starting in existing Tokio runtime
2023-10-08T17:33:05.374124Z  INFO mina_indexer::server: Starting mina-indexer server
2023-10-08T17:33:05.374144Z  INFO mina_indexer::server: Initializing indexer state from blocks in /Users/tbernard/.mina-indexer/startup-blocks
```

jstavh:
I'm going to take a look!

trevorbernard:
No it's not fixed yet but should be a relatively straight forward one.

jstavh:
Yes, I was about to write that I tested on my machine and it hangs still:
```
./target/release/mina-indexer server cli \
    --initial-ledger tests/ledgers/mainnet-genesis.json \
    --is-genesis-ledger \
    --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ
2023-10-11T00:47:25.357525Z  INFO actix_server::builder: starting 8 workers
2023-10-11T00:47:25.357586Z  INFO actix_server::server: Tokio runtime found; starting in existing Tokio runtime
2023-10-11T00:47:25.357818Z  INFO mina_indexer::server: Local socket listener started
2023-10-11T00:47:25.358207Z  INFO mina_indexer::server: Starting mina-indexer server
2023-10-11T00:47:25.358218Z  INFO mina_indexer::server: Initializing indexer state from blocks in /Users/.mina-indexer/startup-blocks

Issue #221, opened 2023-10-08
Problem: There is no way to signal the indexer to shutdown


trevorbernard:
Add a shutdown command to the `mina-indexer client` CLI API.

PR #231, opened 2023-10-09
Open IPC Sooner

jenr24G:
Changes the server IPC listener to open before the server initialization, so the client can communicate with the server during initialization. Also adds a test to verify this functionality. Related: #225, #221

### DIFF
--- a/download_blocks
+++ b/download_blocks
@@ -1,9 +1,10 @@
 #!/bin/sh
 set -ex
 
-touch $TMP/block_urls
+TMPDIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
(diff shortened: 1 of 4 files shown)


trevorbernard:
I haven't started the review yet but now that we open the domain socket right away, that also means it's going to accept updates from the file watcher. Is this a problem?

jenr24G:
> I haven't started the review yet but now that we open the domain socket right away, that also means it's going to accept updates from the file watcher. Is this a problem?

No, the file watcher is spawned by the `run()` method. I extracted the IPC code from the `select!` in the `run()` method and moved it to it's own task which I spawn just before calling `initialize()`

trevorbernard (approved):

Issue #232, opened 2023-10-10
Problem: a lot of the development experience is manual


trevorbernard:
Things like executing docker builds, running regression suites, linters are all executed manually. Use just to create a better developer user experience.

AC:
* Add just support to nix
* Add docker build to justfile
* Add running unit tests to justfile
* Add linters to justfile
* Add regression suite to justfile

PR #234, opened 2023-10-10
Add just support

trevorbernard:
* Fix https://github.com/Granola-Team/mina-indexer/issues/232 - Add support for Just

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -1,0 +1,35 @@
+# Justfile
+#
+# The command 'just' will give usage information.
+# See https://github.com/casey/just for more.
+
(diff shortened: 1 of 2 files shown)


robinbb:
Would prefer that "support for Just" be added before "using Just in CI" in another PR.

trevorbernard:
> Would prefer that "support for Just" be added before "using Just in CI" in another PR.

Ok

robinbb (approved):
Now that I understand what is happening in this PR, I realize that I should not have asked to split the original one into two.

Issue #239, opened 2023-10-11
Problem: There is a bug in the genesis ledger to ledger conversion code for balances


trevorbernard:
Consider the following entry in the genesis ledger:
```javascript
{
    "pk": "B62qqDJCQsfDoHJvJCh1hgTpiVbmgBg8SbNKLMXsjuVsX5pxCELDyFk",
    "balance": "148837.2",
    "delegate": "B62qp8Vq6n4VHq1LUm9Wd5QKjpKb7umoZ2oU9gpJYuHNUc7t2HGhGUA",
    "timing": {
        "initial_minimum_balance": "148837.2",
        "cliff_time": "86400",
        "cliff_amount": "148837.2",
        "vesting_period": "1",
        "vesting_increment": "0"
}
```
Start an indexer with no blocks
```bash
target/release/mina-indexer server cli --initial-ledger ~/mainnet.json -s ~/.mina-indexer/blocks/empty
```
Query the account balance for the above public key

```bash
target/release/mina-indexer client account -p B62qqDJCQsfDoHJvJCh1hgTpiVbmgBg8SbNKLMXsjuVsX5pxCELDyFk
{
  pk:       B62qqDJCQsfDoHJvJCh1hgTpiVbmgBg8SbNKLMXsjuVsX5pxCELDyFk
  balance:  0
  nonce:    0
  delegate: B62qp8Vq6n4VHq1LUm9Wd5QKjpKb7umoZ2oU9gpJYuHNUc7t2HGhGUA
}
```
The balance should be `148837.2`

jenr24G:
got a fix ready!

trevorbernard:
Resolved with the following PR: https://github.com/Granola-Team/mina-indexer/pull/240

PR #240, opened 2023-10-11
Genesis Parsing Fix

jenr24G:
The Mina Indexer parses account balances from the genesis ledger incorrectly. Mina genesis ledgers are denominated in regular mina, while the Mina Indexer expects NanoMina.

Solves https://github.com/Granola-Team/mina-indexer/issues/239

This PR Makes the following changes:
* Parse the account balance on genesis ledger json files using the `Decimal` type
* Panic if the account balance doesn't exist or can't be converted to NanoMina (indicating a faulty genesis ledger)


trevorbernard (changes requested):
Add a unit or regression test that proves it's been fixed

test
```diff
@@ -124,4 +124,18 @@ idxr server cli \
...
+dl_mainnet 5 $TMP/blocks
+server cli \
+    --startup-dir $TMP/blocks \
+    --watch-dir $TMP/blocks \
+    --initial-ledger tests/ledgers/mainnet-genesis.json \
+    --is-genesis-ledger \
+    --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ &
+IDXR_PID=$(pgrep mina-indexer)
+sleep 3
+echo $(client -o account --public-key B62qqDJCQsfDoHJvJCh1hgTpiVbmgBg8SbNKLMXsjuVsX5pxCELDyFk) |
```
trevorbernard, line 137: You don't need the subshell here.

```
client -o account --public-key B62qqDJCQsfDoHJvJCh1hgTpiVbmgBg8SbNKLMXsjuVsX5pxCELDyFk |
```

test
```diff
@@ -124,4 +124,18 @@ idxr server cli \
 sleep 3
 shutdown
 
+# Indexer server reports correct balance for Genesis Ledger Account
+dl_mainnet 5 $TMP/blocks
```
trevorbernard, line 128: You don't need to download any blocks since it's a genesis ledger test

trevorbernard (changes requested):

test
```diff
@@ -124,4 +124,18 @@ idxr server cli \
...
+server cli \
+    --startup-dir $TMP/blocks \
+    --watch-dir $TMP/blocks \
+    --initial-ledger tests/ledgers/mainnet-genesis.json \
+    --is-genesis-ledger \
+    --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ &
+IDXR_PID=$(pgrep mina-indexer)
+sleep 3
```
trevorbernard, line 136: The CI is a bit slow so I would sleep 5 so the regression tests pass

trevorbernard (changes requested):

trevorbernard (approved):

PR #241, opened 2023-10-11
Addressing issue #225

jstavh:
Addressing `Problem: the indexer server isn't correctly initialized if the startup-blocks directory doesn't exist #225` https://github.com/Granola-Team/mina-indexer/issues/225

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -279,7 +279,7 @@ pub async fn initialize(
         ledger,
         is_genesis_ledger,
         root_hash,
-        startup_dir: _,
+        startup_dir,
         watch_dir: _,
         prune_interval,
         canonical_threshold,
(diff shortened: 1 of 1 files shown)


trevorbernard:
@jstavh The solution is to create the `startup_dir` if it doesn't exist. There is a function in `server.rs` called `create_dir_if_non_existent`


You call it like this
```
create_dir_if_non_existent(startup_dir).await;
```

jstavh:
> @jstavh The solution is to create the `startup_dir` if it doesn't exist. There is a function in `server.rs` called `create_dir_if_non_existent`
> 
> You call it like this
> 
> ```
> create_dir_if_non_existent(startup_dir).await;
> ```

@trevorbernard thanks! Just seeing this and will work in this direction.

trevorbernard (approved):

PR #245, opened 2023-10-12
Refactor regression tests to be their own functions and general improvements

trevorbernard:
Add the ability to run individual tests.

`./test` runs all the tests
`./test test_best_canonical_tip test_indexer_cli_reports` would only run those two tests

### DIFF
--- a/test
+++ b/test
@@ -5,11 +5,11 @@ DOMAIN_SOCKET="$(pwd)/mina-indexer.sock"
 
 TMPDIR=$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')
 BLOCKS_DIR="${TMPDIR}/blocks"
-mkdir $BLOCKS_DIR
+mkdir "${BLOCKS_DIR}"
 
 cleanup() {
     echo "Exiting with $?"
-    rm -rf $TMPDIR 2>/dev/null
+    rm -rf ${TMPDIR} 2>/dev/null
 }
 
 trap cleanup EXIT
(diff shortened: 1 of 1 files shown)


jenr24G (approved):
good stuff. maybe we can start splitting test cases off into their own files after this

test
```diff
@@ -3,135 +3,154 @@ set -ex
...
-    rm -rf $TMPDIR
-}
-trap cleanup EXIT
-
-cargo build
-
-teardown() {
-    rm -rf $HOME/.mina-indexer/database || true
-    rm -rf $HOME/.mina-indexer/logs || true
+    rm -rf $TMPDIR 2>/dev/null
```
robinbb, line 12: Could be disastrous if $TMPDIR is a directory with a space character in any of the directory components. Recommend surrounding by double quotes. Safer still is "${TMPDIR}".

test
```diff
@@ -3,135 +3,154 @@ set -ex
...
+setup() {
+    echo "setup"
 }
 
-client() {
-    ./target/debug/mina-indexer client "$@"
+teardown() {
+    echo "Cleaning up and killing the background indexer with PID: $idxr_pid"
+    kill $idxr_pid 2>/dev/null
+    rm -rf $HOME/.mina-indexer/database || true
```
robinbb, line 45: Same comment as above applies with this line and the next one.

robinbb (dismissed):
Approved based on C4 rules, but see my comments about use of 'rm -fr' on an unquoted variable.

test
```diff
@@ -3,135 +3,155 @@ set -ex
...
-    rm -rf $TMPDIR
-}
-trap cleanup EXIT
-
-cargo build
-
-teardown() {
-    rm -rf $HOME/.mina-indexer/database || true
-    rm -rf $HOME/.mina-indexer/logs || true
+    rm -rf ${TMPDIR} 2>/dev/null
```
robinbb, line 12: To be clear, it's the surrounding double quotes, not the surrounding braces, that save you from having the TMPDIR being interpreted as multiple directories. So, this is still not safe, FYI.

robinbb (approved):

Issue #251, opened 2023-10-19
Problem: Regression suite does not clean up `$USER/.mina-indexer/database` after Tests


jenr24:
The Mina Indexer does not currently clean up its rocksdb journal after a test has executed, leading to multiple executions of the testing suite failing with a `LOCK` in the database dir

PR #253, opened 2023-10-23
Fix Test Bugs

### LINKED ISSUE #252: Problem: regression tests do not pass locally on main
on the latest commit for main (05f5221), the regression tests do not pass on my local machine. the consistently failing test is `test_best_canonical_tip` which is failing with the error `Error: invalid type: string "No summary available yet", expected struct SummaryShort at line 1 column 26`

jenr24G:
This pull request fixes the inconsistencies with running the regression test suite locally, as well as some QoL changes regarding removing dangling socket files, etc.

Closes #252, #251

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -7,10 +7,10 @@ use crate::{
         IndexerState, Tip,
     },
     store::IndexerStore,
-    MAINNET_TRANSITION_FRONTIER_K, SOCKET_NAME,
+    MAINNET_TRANSITION_FRONTIER_K, SOCKET_NAME, ipc::IpcActor,
 };
 use anyhow::anyhow;
-use futures::io::{AsyncBufReadExt, AsyncWriteExt, BufReader};
+use futures::{io::{AsyncBufReadExt, AsyncWriteExt, BufReader}};
 use interprocess::local_socket::tokio::{LocalSocketListener, LocalSocketStream};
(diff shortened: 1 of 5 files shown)


src/server.rs
```diff
@@ -295,26 +232,32 @@ pub async fn initialize(
...
             root_hash.clone(),
-            ledger.ledger,
+            ledger.ledger.clone(),
             store,
             MAINNET_TRANSITION_FRONTIER_K,
             prune_interval,
             canonical_update_threshold,
         )?;
 
+        info!("Getting best tip");
```
trevorbernard, line 242: These should be debug. Non-blocking

src/server.rs
```diff
@@ -350,11 +289,11 @@ pub async fn run(
     block_watch_dir: impl AsRef<Path>,
     mut state: IndexerState,
     phase_sender: Arc<watch::Sender<MinaIndexerRunPhase>>,
-    mut query_receiver: mpsc::Receiver<(
+    mut query_receiver: mpsc::UnboundedReceiver<(
```
trevorbernard, line 292: Using an unbounded array for IPC is dubious. I would prefer to see it bounded but something more reasonable than `50` like `1024`. Non-blocking

trevorbernard (approved):

Issue #237, opened 2023-10-11
Problem: Google Cloud Block Receiver is unused


trevorbernard:
Remove the Google Cloud module from the codebase as it's unused. Also remove unused dependencies in `Cargo.toml` caused by removing the Google Cloud module.

jenr24:
sad day this was part of the code I wrote for Vitor to merge Iggy into the main codebase. I can take this!

PR #254, opened 2023-10-23
Remove google cloud as it's incomplete and unused

trevorbernard:
This PR fixes: https://github.com/Granola-Team/mina-indexer/issues/237


jenr24G (approved):

robinbb (approved):
We can always recover this using the Git history.

PR #261, opened 2023-10-24
Add Buildkite pipeline

robinbb:
Squash merge, please. This PR adds a Buildkite pipeline and therein reproduces all functionality of the existing GitHub actions, except for the unit tests and the image publishing, which remain active.


ops/test-justfile-targets
```diff
@@ -0,0 +1,13 @@
+#! /bin/sh
+
+set -ex
+
+just
+just clean
+just audit
+just lint
+just build
+# just test-unit
```
trevorbernard, line 10: Should remove commented code. Non-blocking

ops/test-justfile-targets
```diff
@@ -0,0 +1,13 @@
...
+just test-regression
+# just test
+# just test-ci
```
trevorbernard, line 13: Same as above

test
```diff
@@ -3,19 +3,19 @@ set -ex
...
-    rm -rf ${TMPDIR} 2>/dev/null
-    rm -rf "${HOME}/.mina-indexer/database" || true
-    rm -rf "${HOME}/.mina-indexer/logs" || true
-    rm ${DOMAIN_SOCKET} || true
+test_cleanup() {
+    echo "Test result: $?"
+    rm -rf "${TMPDIR}"
+    rm -rf "${HOME}/.mina-indexer/database"
+    rm -rf "${HOME}/.mina-indexer/logs"
+    rm -rf "${DOMAIN_SOCKET}"
```
trevorbernard, line 15: We should ignore the exit code. If the domain socket gets handled by the indexer this will fail

trevorbernard (approved):

PR #270, opened 2023-10-29
Use temporary directories for logs, database

### LINKED ISSUE #268: Problem: CI runs are polluting $HOME/.mina-indexer and interfering with each other

### LINKED ISSUE #271: Problem: socket is shared among indexer instances
`mina-indexer.sock` is hard-coded, so if a user starts 2+ instances of the indexer from within the same directory (a likely scenario in CI), then there is confusion because there is only 1 'mina-indexer.sock' in that directory. How can it be shared among the various instances, reasonably?

robinbb:
Fixes #268
Fixes #271

### DIFF
--- a/test
+++ b/test
@@ -1,7 +1,7 @@
 #!/bin/sh
 set -ex
 
-DOMAIN_SOCKET='mina-indexer.sock'
+# Collect the binaries under test and the test ledger.
 IDXR="$(pwd)"/target/debug/mina-indexer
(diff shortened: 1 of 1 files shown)


test
```diff
@@ -134,9 +154,12 @@ test_account_balance_cli() {
 
 test_account_public_key_json() {
     setup
-    mkdir "$BLOCKS_DIR"/database
```
robinbb, line 137: Addresses your comment from prior PR, @trevorbernard

trevorbernard (approved):

Issue #288, opened 2023-11-01
Problem: the `mina-rs` package has dependencies that are severe security risks


robinbb:
... yet we don't really need mina-rs.

sunfishcode:
According to mina-indexer's checked in Cargo.lock, mina-indexer depends on `rustix` via `tempfile` and `which`. I have audited both of those libraries and neither use `rustix::fs::Dir` anywhere. So, mina-indexer is unaffected by [GHSA-c827-hfw6-qwvm](https://github.com/bytecodealliance/rustix/security/advisories/GHSA-c827-hfw6-qwvm).

trevorbernard:
Even though we weren't affected, an update to the dependencies resolved this issue. Closing https://github.com/Granola-Team/mina-indexer/pull/289

robinbb:
Appreciate that audit, @sunfishcode.

PR #289, opened 2023-11-01
Update Rust package dependencies

robinbb:
Removes doubt about whether updating dependencies will resolve problems with the use of (transitive) unsafe/insecure dependencies. It won't the problems, it seems.


trevorbernard (approved):

trevorbernard:
This fixes: https://github.com/Granola-Team/mina-indexer/issues/288

Issue #283, opened 2023-10-31
Problem: some CLI arg descriptions are incorrect


trevorbernard:
Solution: correct the the bad descriptions

trevorbernard:
It's resolved

Issue #269, opened 2023-10-29
Problem: "mina-indexer server cli" is a rididulous command line


robinbb:
"server" and "CLI" are not the same thing, yet we require both on the command line to launch what? A server? A CLI? A client server? A server client? This is bad usability.

trevorbernard:
Agreed

trevorbernard:
This issue has been resolved

PR #302, opened 2023-11-05
Improve cli commands

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/283, https://github.com/Granola-Team/mina-indexer/issues/269

Improving the ergonomics of the CLI interface phase 1.

### DIFF
--- a/src/bin/mina-indexer.rs
+++ b/src/bin/mina-indexer.rs
@@ -26,7 +26,6 @@ struct Cli {
 enum IndexerCommand {
     /// Server arguments
     Server(ServerArgs),
-
     /// Client commands
     Client {
         /// Output JSON data when possible
--- a/src/bin/mina-indexer.rs
+++ b/src/bin/mina-indexer.rs
@@ -24,11 +24,9 @@ struct Cli {
 
 #[derive(Subcommand, Debug)]
 enum IndexerCommand {
-    /// Server commands
-    Server {
-        #[command(subcommand)]
-        server_command: ServerCommand,
-    },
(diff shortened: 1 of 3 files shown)


src/bin/mina-indexer.rs
```diff
@@ -39,25 +36,14 @@ enum IndexerCommand {
     },
 }
 
-#[derive(Subcommand, Debug)]
-enum ServerCommand {
-    /// Start the mina indexer with a config file
-    Config {
```
robinbb, line 45: Here you are removing functionality that was indeed tested (to some minimal degree). Do we support a config file or no? If no, then surely there is some sort of YAML parsing code that should be deleted. If yes, then how do we specify where the config file is located?

robinbb (changes requested):
The feature of parsing a YAML config file is deleted with this PR. It is not clear to me that we agree that we should not parse a config file, so I suggest splitting this PR into 2: the CLI changes, and then the config file support deletion.

src/bin/mina-indexer.rs
trevorbernard: The functionality is not easily replicated. There was a reason why the additional CLI command was used in favour of command line/config file precedence. For sake of change latency I was going to create another issue to add it back in.

robinbb: Okay, if you create that issue, refer to it here, and then merge in current 'main', I'll approve and merge.

trevorbernard:
@robinbb Ok, I'll make those changes

PR #309, opened 2023-11-06
Decrease build pipeline steps

### LINKED ISSUE #310: Problem: CI does not cache intermediate build output between build steps
This is because separate Buildkite "Steps" are, by default, indeed separate and not meant to share stuff.

robinbb:
Fixes #310


robinbb:
Gets build down to 4m.

trevorbernard (approved):

Issue #311, opened 2023-11-06
Problem: `create_dir_if_non_existent` unnecessarily async 


trevorbernard:
Refactor to use std i/o

PR #312, opened 2023-11-06
Replace `create_dir_if_non_existence` with `std::fs::create_dir_all`

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/311

### DIFF
--- a/test
+++ b/test
@@ -172,6 +172,39 @@ test_account_public_key_json() {
     teardown
 }
 
+assert_directory_exists() {
+    directory="$1"
+
+    if [ ! -d "$directory" ]; then
+        echo "Test Failed: Expected directory $directory to exist, but it does not."
+        exit 1
+    else
+        echo "Test Passed: Directory $directory exists."
+    fi
+}
+
+test_startup_dirs_get_created () {
+    setup
+
+    idxr server cli \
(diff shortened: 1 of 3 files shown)


test
```diff
@@ -172,6 +172,39 @@ test_account_public_key_json() {
...
+        --is-genesis-ledger \
+        --root-hash 3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ &
+    sleep 5
+
+    assert_directory_exists "./startup-blocks"
+    assert_directory_exists "./watch-blocks"
+    assert_directory_exists "./database"
+    assert_directory_exists "./logs"
+
+    idxr_pid=$(pgrep mina-indexer)
```
robinbb, line 204: At some point, we need a better way to get the PID because there could be multiple indexers running. Maybe we write to the current directory a PID file?

robinbb (approved):
Nice to hack away at the unnecessary async stuff.

test
trevorbernard: Agreed. We'll need something that abstracts away and isolates the system under test. We'll have to deal with a webserver port soon enough too.

Issue #323, opened 2023-11-10
Problem: `is_valid_block_file` can return false positives


trevorbernard:
Support only these two patterns explicitly:

```rust
// <network>-<block_height>-<state_hash>.json
// <network>-<state_hash>.json
```
Ensure block_height is a u64, the extension is `json` and the file_name matches the pattern

trevorbernard:
This would be a excellent candidate for creating a property based test

Isaac-DeFrain:
Blockchain_length should be `u32`

PR #347, opened 2023-12-30
Improve file name validation

Isaac-DeFrain:
- add property-based testing for block file names

Fixes https://github.com/Granola-Team/mina-indexer/issues/323

### DIFF
--- a/src/block/mod.rs
+++ b/src/block/mod.rs
@@ -166,13 +166,13 @@ impl std::fmt::Debug for BlockHash {
     }
 }
 
-pub fn parse_file(filename: &Path) -> anyhow::Result<PrecomputedBlock> {
-    if is_valid_block_file(filename) {
-        let blockchain_length =
-            get_blockchain_length(filename.file_name().expect("filename already checked"));
-        let state_hash = get_state_hash(filename.file_name().expect("filename already checked"))
-            .expect("state hash already checked");
-        let log_file_contents = std::fs::read(filename)?;
+/// Parses the precomputed block if the path is a valid block file
+pub fn parse_file(path: &Path) -> anyhow::Result<PrecomputedBlock> {
+    if is_valid_block_file(path) {
+        let file_name = path.file_name().expect("filename already checked");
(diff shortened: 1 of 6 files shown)


src/block/mod.rs
```diff
@@ -182,24 +182,19 @@ pub fn parse_file(filename: &Path) -> anyhow::Result<PrecomputedBlock> {
...
 pub fn get_state_hash(file_name: &OsStr) -> Option<String> {
     let last_part = file_name.to_str()?.split('-').last()?.to_string();
-    if last_part.starts_with('.') || !last_part.starts_with("3N") {
-        return None;
-    }
-
     let state_hash = last_part.split('.').next()?;
-    if state_hash.contains('.') {
-        return None;
+    if state_hash.starts_with("3N") {
```
n1tranquilla, line 194: Significance of "3N" ?

src/block/mod.rs
```diff
@@ -214,13 +209,47 @@ pub fn get_blockchain_length(file_name: &OsStr) -> Option<u32> {
...
+    if let Some(ext) = path.extension() {
+        // check json extension
+        if ext.to_str() == Some("json") {
+            // check file stem
+            if let Some(file_name) = path.file_stem() {
+                if let Some(parts) = file_name
+                    .to_str()
+                    .map(|name| name.split('-').collect::<Vec<&str>>())
+                {
+                    if parts.len() == 2 {
```
n1tranquilla, line 224: is there anyway use `.last()` for `Vec`? Might simplify this. I'm not blocking the MR on this.

src/block/mod.rs
Isaac-DeFrain: Just some bytes to distinguish state hashes from the other various types of hashes used

n1tranquilla: Might it be helpful to capture the type of hash in a variable?

src/block/mod.rs
Isaac-DeFrain: Great suggestion, much simpler that way

trevorbernard (approved):

PR #397, opened 2024-01-26
Convert from RocksDB to Speedb

robinbb:
Tried this to see if it helps build on Mac. It does not. The same compiler error arises because Speedb forked the RocksDB code, and the same file ('repair.cc') causes a compilation failure. However, this might result in a faster indexer. We should merge, if so.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 ## [Unreleased]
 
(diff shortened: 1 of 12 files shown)


Isaac-DeFrain (approved):
Merging optimistically. We should do some perf testing with rocksdb and speeddb

PR #426, opened 2024-01-31
Use release profile and omit unnecessary check

### LINKED ISSUE #418: Problem: CI is slower than it needs to be
Between building, clippy, checks, audits and testing, the codebase gets rebuilt more than necessary because of profiles. Make sure they all build to the same target. Also, cargo check is redundant when clippy is also being run. That can be removed.

robinbb:
Fixes #418


robinbb:
@Isaac-DeFrain I need help on this. Are the unit tests passing because they are incorrect? I think so. The alternative is that the Rust compiler generates incorrect code when using the 'release' profile, which I think unlikely.

trevorbernard (approved):

PR #447, opened 2024-02-07
Have rustup manage rust-analyzer

trevorbernard:
Remove the explicit dependency on rust-analyzer in flake.nix in favour of using rust-toolchain.toml to manage it. So for those who do not use nix will have the same version of rust-analyzer.


Isaac-DeFrain (approved):
LGTM

PR #477, opened 2024-02-15
Improve canonical chain discovery block sorting

trevorbernard:
Sorting all mainnet blocks previously took 13hrs and it's not down to 100ms


Isaac-DeFrain (approved):
Super fast!

PR #484, opened 2024-02-20
Dramatically improve ingestion performance in canonical chain discovery

trevorbernard:
* Improve performance of extracting block height and state hash from filenames
* Add a canonical state hashes set for improved performance when getting orphaned paths

### DIFF
--- a/src/canonicity/canonical_chain_discovery.rs
+++ b/src/canonicity/canonical_chain_discovery.rs
@@ -1,8 +1,8 @@
 use crate::block::{
-    blockchain_length::*, extract_block_height, extract_block_height_or_max, extract_state_hash,
-    previous_state_hash::*,
+    extract_block_height, extract_block_height_or_max, extract_state_hash, previous_state_hash::*,
 };
 use std::{
+    collections::HashSet,
     path::{Path, PathBuf},
     time::Instant,
 };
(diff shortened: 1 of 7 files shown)


robinbb:
Was this really intended to remain in draft?

trevorbernard:
> Was this really intended to remain in draft, @trevorbernard ?

Yes, I wanted to rebase those nix changes in and @Isaac-DeFrain to take a look at my changes. I changed one assumption in the code and I wanted to verify it's validity before I truly PR'ed it.

src/canonicity/canonical_chain_discovery.rs
```diff
@@ -205,24 +191,20 @@ pub fn discovery(
...
+
     let orphaned: Vec<PathBuf> = paths
         .into_iter()
         .filter(|p| {
-            if let Ok(length) = BlockchainLength::from_path(p) {
-                let length: u32 = length.into();
-                return length <= max_canonical_length && !canonical_paths.contains(p);
+            if let Some(length) = extract_block_height(p) {
+                return length <= max_canonical_length
+                    && !canonical_state_hashes.contains(&extract_state_hash(p));
```
trevorbernard, line 202: Seems to me we could eliminate the block height check and solely rely on the canonical state hashes set. Since orphaned paths are the set difference between all paths and those that are canonical.

Isaac-DeFrain: We're doing the length check because it's not a strict set difference. We have 3 collections:
- `canonical_paths` (up to best block height - canonical threshold)
- `successive_paths` (subtree rooted at the highest block in `canonical_paths`)
- `oprhaned_paths` (the remaining paths)

Isaac-DeFrain: Names still need to be changed

Isaac-DeFrain (approved):
LGTM

PR #486, opened 2024-02-20
(canonicity) Generate correct orphaned paths

trevorbernard:
There is a regression with the recent performance enhancements. Some precomputed blocks do no contain the block height in filename. In that case, parse the height from the precomputed json block.


trevorbernard:
Closing PR as it's not valid

Issue #342, opened 2023-12-05
Problem: `mina-rs` project is unmaintained but we rely on the `mainnet` block parsing code


trevorbernard:
Create hermetic parsing code that replaces the `mina-serialization` module in `mina-rs`

Isaac-DeFrain:
@jhult will take over

Isaac-DeFrain:
- `src/mina_blocks/v1` has all the block (de)serializer bits
- we will need to implement their `version` lib to use for transaction hashing (one can see what a versioned block looks like by using `mina-indexer best-tip -v` or any other verbose block query)
- a lot of our internal structs use mina-rs structs but should only required a mechanical change to use ours

PR #507, opened 2024-02-26
Remove mina-rs dependency

jhult:
Fixes https://github.com/Granola-Team/mina-indexer/issues/342

I have not yet explored the overlap with `mina_blocks`. That will be handled in a second PR.


trevorbernard:
@jhult I wonder if we can't turn off clippy for the mina-rs code

.gitignore
```diff
@@ -1,9 +1,204 @@
...
+*.cab
+*.msi
+*.msix
+*.msm
+*.msp
+
+# Windows shortcuts
+*.lnk
+
+# End of https://www.toptal.com/developers/gitignore/api/rust,linux,macos,windows,intellij+iml,rust-analyzer,visualstudiocode
```
trevorbernard, line 204: This `.gitignore` doesn't understand our application on some of it wrong. Like removing the previous entries and ignoring things like `Cargo.lock`. I would revert back to the previous `.gitignore` file and add only what needs to be ignored. E.g. Intellij project files

trevorbernard (changes requested):

trevorbernard:
Also, I would also update documentation that refers to git submodules. Like in the readme for nix.

```
nix build .?submodules=1
nix build
```

And I would remove `git submodule update --init --recursive` from the shell hook

.gitignore
jhult: I will revert this.

FWIW, I tend to use `.gitignore` files based on [toptal.com/developers/gitignore](https://www.toptal.com/developers/gitignore/). However, these aren't perfect and often need tweaks depending on the project.

trevorbernard (approved):

PR #508, opened 2024-02-26
Remove proof so serde_json::from_slice(&bytes) works

jhult:
Original: https://github.com/Granola-Team/mina-rs/pull/3

### DIFF
--- a/src/protocol/serialization_types/snark_work.rs
+++ b/src/protocol/serialization_types/snark_work.rs
@@ -19,7 +19,6 @@ use serde::{Deserialize, Serialize};
 pub struct TransactionSnarkWork {
     // Versioned 1 byte
     pub fee: AmountV1,
-    pub proofs: OneORTwoV1,
     pub prover: PublicKeyV1,
 }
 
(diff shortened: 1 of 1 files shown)


trevorbernard:
This needs more discussion as the change will likely break something inside mina-rs serialization.

trevorbernard:
The primary reason for this breaking change is to improve the ingestion performance of the Precomputed Blocks. Unfortunately the JSON in the PCB is not valid JSON as it's not valid UTF8 unicode. Most of this non-compliance lies in the various proofs which are unnecessary for our block ingestion.

So as a work around when parsing PCBs we must read all the bytes and call `String::from_utf8_lossy(...)`. As the name implies, it replaces all invalid utf8 characters with the following: https://doc.rust-lang.org/std/char/constant.REPLACEMENT_CHARACTER.html. This is a bandaid to get the PCB to deserialize but is also inefficient as we need to allocate a large string and do a cleaning pass.

With this change we can use the much more efficient `serde_json::from_slice(&bytes)` with slight modification to our internal PCB. This enabled me to ingest 190k blocks in 43mins.

robinbb:
@jhult @trevorbernard What is the plan with this PR?

robinbb:
@jhult I am doing an enormous amount of block ingestion right now, in an effort to ingest the entire chain. If there is a performance increase to be had with this work, then now is the time to invest in it. Is there an easy win, here?

trevorbernard:
Closing this PR and I'll make a proper one with all the fixin's

Issue #509, opened 2024-02-26
Enhancement: Merge Rust imports using rustfmt


jhult:
Use `imports_granularity = "Crate"` [in rustfmt](https://rust-lang.github.io/rustfmt/?search=&version=v1.6.0#imports_granularity) to merge Rust imports (for cleaner imports).

### What this does?

<img width="919" alt="image" src="https://github.com/Granola-Team/mina-indexer/assets/9849069/e9217da0-dc25-4460-ad3d-45574858cce5">

### Example
<img width="673" alt="image" src="https://github.com/Granola-Team/mina-indexer/assets/9849069/083e50dc-d5a9-4aa4-9807-a9d8e7a3d781">

trevorbernard:
We should create a `rustfmt.toml` file for this. I would like to see comments wrapped to 80 chars

PR #540, opened 2024-03-06
Fix rustfmt in nix, remove duplicate clippy command, format files

jhult:
This is a follow on to Granola-Team/mina-indexer#509.

I had to add in formatting of all the files so it would pass the formatting checks in CI. I did this in a separate commit (for ease of review).


jhult:
FYI, I think if @Isaac-DeFrain cherry-picked [this commit](https://github.com/Granola-Team/mina-indexer/pull/540/commits/15db9cd511c9be710abc9bc71c6a8ce6b7b7e463), he could then run `just format`, and it would format all his files. But, there would likely still be conflicts after. I should probably just drop the commit formatting all the code and re-do it after rebasing on any changes from @Isaac-DeFrain.

robinbb:
This "holding this PR for the upcoming big PR to avoid conflicts on the big PR" pattern is an antipattern. It fails to disincentivize the bad behaviour (big PRs), and threatens to "spread" the problem (unshipped value) to other PRs (by not shipping the value in this PR). Also, it is contrary to C4. The solution is to never do that. I will merge this. Conflicts can be avoided on the big PR in the way that @jhult explains.

robinbb (approved):

PR #575, opened 2024-03-08
Modify justfile to use nightly toolchain for fmt and check

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/574
* https://github.com/Granola-Team/mina-indexer/issues/574


trevorbernard:
This breaks in the nix environment

PR #600, opened 2024-03-13
Ensure rustfmt works in all environments

jhult:
Fixes Granola-Team/mina-indexer#574

Nix environment has `rustfmt` nightly and won't work with `+nightly`. Non-Nix environment needs nightly toolchain installed and requires `+nightly`.

I ended up checking the `rustfmt` version (`stable` or `nightly`) instead of checking to see if we are inside the Nix environment. However, I'm leaving related reading about how to determine if inside a Nix environment: https://github.com/NixOS/nix/issues/6677

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -3,6 +3,12 @@
 # The command 'just' will give usage information.
 # See https://github.com/casey/just for more.
 
+# Ensure rustfmt works in all environments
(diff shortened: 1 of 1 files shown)


Isaac-DeFrain (approved):
LGTM

Issue #615, opened 2024-03-15
Problem: unable to build docker images from nix


trevorbernard:
Add build docker image support to nix

Issue #619, opened 2024-03-17
Problem: The indexer doesn't have a concept of a `production` release


trevorbernard:
Follow the [Filesystem Heirachy Standard](https://refspecs.linuxfoundation.org/FHS_3.0/fhs/index.html) for production releases.

- locked.cvs, genesis block and ledger should be read from `/usr/share/mina-indexer/data` 
- Indexer database stored in `/var/lib/mina-indexer/database`
- mina-indexer binary in `/usr/bin`

trevorbernard:
Maybe we should use something along the lines of https://docs.rs/clap_conf/latest/clap_conf/ to unify the CLI args, env variables and config files?

trevorbernard:
I think we need to set default values conditionally in `clap`. Production values point to the standard FHS otherwise maintain status quo.

robinbb:
@trevorbernard I think that the 'nix build' generates a package that does indeed follow the FHS, and is usable as a production release. Are there any remaining directories or files to add? Here is what I have:

robinbb:
```
|-- bin
|   `-- mina-indexer
|-- usr
|   `-- share
|       `-- mina-indexer
|           `-- data
|               |-- locked.csv
|               |-- mainnet-1-3NKeMoncuHab5ScarV5ViyF16cJPT4taWNSaTLS64Dp67wuXigPZ.json
|               `-- mainnet.json
`-- var
    |-- lib
    |   `-- mina-indexer
    `-- log
        `-- mina-indexer
```

PR #627, opened 2024-03-19
Nix docker image

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/615
Partially work toward: https://github.com/Granola-Team/mina-indexer/issues/619

This refactor is in an effort to better isolate the rust code from the other non source files. The rust code was moved into it's own directory so modifying configuration files like flake.nix wouldn't trigger a rebuild of the mina-indexer derivation.

I added dynamic default values based on the ENV['RELEASE'] variable. This is designed obsolescence. Ideally, this would just set the default value at compile time rather than a runtime dynamic value. But that is an issue for a future time.


trevorbernard:
I cancelled the previous same PR because the build info wasn't showing because of a 502 error. @robinbb or @Isaac-DeFrain will need to approve again

PR #663, opened 2024-03-25
Deprecate nix-shell

robinbb:


Isaac-DeFrain (approved):
LGTM

Justfile
```diff
@@ -34,8 +34,7 @@ format:
 test: test-unit
   cd rust && ./test
 
-test-ci: lint test-unit
-  cd rust && ./test
+test-ci: lint test-unit test-regression
```
jhult, line 37: If `test-unit` and `test-regression` both depend on `build`, won't this try to build twice and slow CI down?

robinbb: Only slightly. The second invocation of 'cargo' should find that there's nothing to build. (Optimally, the build toot itself - 'just' - would discover this and then not even invoke the second 'cargo', but that's not how 'just' works. A Makefile would be better at this.)

robinbb: FYI, I made this change so that the docs (which I changed due to nix-shell disappearing) had their build target tested in CI. Arguably, that is not related to this PR, and should have been moved to another PR, avoiding this valid discussion on a matter nearly unrelated to nix-shell deprecation. My bad.

Isaac-DeFrain: Resolving to merge

Issue #654, opened 2024-03-22
Problem: we don't index internal commands


Isaac-DeFrain:

trevorbernard:
Store internal commands (fee transfer, coinbase, fee transfer via coinbase) making sure to include it's parent block's state_hash.

Isaac-DeFrain:
User commands are indexed on:
- public keys
- state hash of block
- command/transaction hash

AFAIK internal commands do not have hashes associated with them. We'll index internal commands on:
- public keys
- state hash of block

PR #667, opened 2024-03-26
Store internal commands

Isaac-DeFrain:
- indexes internal commands by public key and state hash

Fixes https://github.com/Granola-Team/mina-indexer/issues/654


robinbb (approved):
Good stuff! Great use of types. Many drive-by improvements are also much appreciated.

Issue #670, opened 2024-03-26
Improvement: Allow specifying Unix domain socket path as global CLI option


jhult:
Related: #650, #636, and #271

Issue #650, opened 2024-03-22
Improvement: refine CLI commands, options, and communication with server 


jhult:
Unflatten commands and simplify communication between CLI and server.

Example:

```sh
Usage: mina-indexer [OPTIONS] <COMMAND>

Commands:
  server                  
  accounts
  blocks
  chain
  checkpoints
  ledger
  staking-ledger
  snark
  summary
  transactions
  help
```

Builds on top of improvements in #648.

jhult:
Closed via #681.

PR #672, opened 2024-03-26
feat: add CLI option for Unix domain socket path and organize shutdown

jhult:
Implements Granola-Team/mina-indexer#670.

Also part of #650.


Isaac-DeFrain (approved):
LGTM

Issue #666, opened 2024-03-26
Problem: `config.json` output to wrong directory on startup


Isaac-DeFrain:
`config.json` (server args json) is incorrectly written to the parent of the log directory

Issue #665, opened 2024-03-25
Problem: we cannot start the indexer from `config.json`


Isaac-DeFrain:
The indexer previously had the ability to start from a config file. That feature was never tested, ultimately it broke and was removed. We should resurrect this feature

jhult:
Related to #666 and #650.

robinbb:
Not sure I agree that this is a problem. This is a feature request to add the ability to provide a JSON file instead of configs to a sever start command, perhaps?

Isaac-DeFrain:
I added some more context above (sorry, I thought I did that before).

robinbb:
I still do not agree that this feature is required. My view is that they existence of the config.json is the problem. See #675

Isaac-DeFrain:
We can remove it

PR #674, opened 2024-03-26
Start from config file

Isaac-DeFrain:
- restores the indexer's ability to start from a config file
- logs config over writing to a file on startup

Fixes https://github.com/Granola-Team/mina-indexer/issues/665
Fixes https://github.com/Granola-Team/mina-indexer/issues/666
Fixes https://github.com/Granola-Team/mina-indexer/issues/675

### DIFF
--- a/rust/src/bin/mina-indexer.rs
+++ b/rust/src/bin/mina-indexer.rs
@@ -37,8 +37,8 @@ enum IndexerCommand {
 enum ServerCommand {
     /// Start a new mina indexer by passing arguments on the command line
     Start(ServerArgs),
-    /// Start a new mina indexer from a config file
-    Config(ConfigArgs),
+    /// Start a new mina indexer via a config file
+    StartViaConfig(ConfigArgs),
     /// Start a mina indexer by replaying events from an existing indexer store
     Replay(ServerArgs),
     /// Start a mina indexer by syncing from events in an existing indexer store
@@ -84,10 +84,6 @@ pub struct ServerArgs {
     /// Path to directory for logs
     #[arg(long, default_value = "/var/log/mina-indexer")]
     pub log_dir: PathBuf,
-
-    /// Path to the server config file
-    #[arg(long)]
-    pub config_file: Option<PathBuf>,
 
     /// Max stdout log level
     #[arg(long, default_value_t = LevelFilter::INFO)]
@@ -192,7 +188,7 @@ pub async fn main() -> anyhow::Result<()> {
                 ServerCommand::Start(args) => (args, InitializationMode::New),
                 ServerCommand::Sync(args) => (args, InitializationMode::Sync),
                 ServerCommand::Replay(args) => (args, InitializationMode::Replay),
-                ServerCommand::Config(args) => {
+                ServerCommand::StartViaConfig(args) => {
                     let contents = std::fs::read(args.path.expect("server args config file"))?;
                     let args: ServerArgsJson = serde_json::from_slice(&contents)?;
                     (args.into(), InitializationMode::New)
@@ -204,6 +200,7 @@ pub async fn main() -> anyhow::Result<()> {
             let web_hostname = args.web_hostname.clone();
             let web_port = args.web_port;
 
+            // default to sync if there's a nonempty db dir
             if let Ok(dir) = std::fs::read_dir(database_dir.clone()) {
                 if matches!(mode, InitializationMode::New) && dir.count() != 0 {
                     // sync from existing db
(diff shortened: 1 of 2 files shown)


rust/test
```diff
@@ -1478,6 +1490,35 @@ test_internal_commands() {
...
+        --database-dir ./database \
+        --log-dir ./logs \
+        --config-file ./logs/config.json
+    wait_for_socket
+    teardown
+
+    setup
+    dl_mainnet 15 ./blocks
+
+    idxr_server config -p ./config.json
```
robinbb, line 1510: So, this means that specifying a config also means "server start" subcommands?

rust/src/bin/mina-indexer.rs
```diff
@@ -198,10 +217,13 @@ pub async fn main() -> anyhow::Result<()> {
...
-            // write server args to config.json
-            let path = log_dir.with_file_name("config.json");
+            // dump server config
+            let path = args
+                .config_file
+                .clone()
+                .unwrap_or(log_dir.with_file_name("config.json"));
             let args_json: ServerArgsJson = args.clone().into();
-            std::fs::write(path, serde_json::to_string_pretty(&args_json)?)?;
+            std::fs::write(&path, serde_json::to_string_pretty(&args_json)?)?;
```
robinbb, line 226: So, this means that we overwrite the supplied config file?

rust/test
Isaac-DeFrain: I'm not 100% set on the name, but those are the semantics I was going for

rust/src/bin/mina-indexer.rs
```diff
@@ -171,13 +185,18 @@ pub async fn main() -> anyhow::Result<()> {
...
         IndexerCommand::Server { server_command } => {
-            let (args, mut mode) = match server_command {
+            let (args, mut mode) = match *server_command {
                 ServerCommand::Shutdown => {
                     return client::run(&client::ClientCli::Shutdown, &domain_socket_path).await;
                 }
                 ServerCommand::Start(args) => (args, InitializationMode::New),
                 ServerCommand::Sync(args) => (args, InitializationMode::Sync),
                 ServerCommand::Replay(args) => (args, InitializationMode::Replay),
+                ServerCommand::Config(args) => {
```
robinbb, line 195: It seems that this is at the wrong level of hierarchy for commands? 'Config' seems to also imply 'start'? Is that intended? If so, maybe this should be called "start-with-config"?

rust/src/bin/mina-indexer.rs
robinbb, line 226: Does this line not overwrite the given config file?

rust/test
```diff
@@ -134,6 +134,18 @@ test_indexer_cli_reports() {
     # Indexer server cli subcommand works with default args
     idxr_server --help 2>/dev/null
 
+    idxr_server start --help 2>&1 |
+        grep -iq "Usage: mina-indexer server start"
+
+    idxr_server config --help 2>&1 |
```
robinbb, line 140: When one "configures" a server, does one also start it?

robinbb (changes requested):
The original problem was that the config.json file is confusing. This issue does not resolve the problem. I am requesting changes based on C4's rule that this does not solve an agreed-upon problem.

rust/src/bin/mina-indexer.rs
Isaac-DeFrain: `start` and `config` are essentially the same; `start` gets CLI args explicitly, `config` gets them from a file

rust/test
Isaac-DeFrain: Yes

Isaac-DeFrain:
@robinbb there are two related issues here
- https://github.com/Granola-Team/mina-indexer/issues/665
- https://github.com/Granola-Team/mina-indexer/issues/666

This PR is intended to solve https://github.com/Granola-Team/mina-indexer/issues/665

rust/test
robinbb: Okay. If those are the semantics, then the name of the switch as-is will confuse users. I recommend changing it to something like "start-with-config-file".

rust/src/bin/mina-indexer.rs
robinbb: This must not be allowed. A first reason is because config files are often installed in a read-only manner. Second, if a sysadmin ever found the modification date on their config file to be other than when they modified it, then they would be astonish or suspicious that their config has been hacked. Third, there is no need to write the file if the configs have not changed. (Can one change the configs on the command line in addition to providing a config file? Often, this is possible, but if so then one does not expect the config file to change.)

rust/test
robinbb: Confusing. See above.

robinbb:
> @robinbb there are two related issues here
> 
> * [Problem: we cannot start the indexer from `config.json` #665](https://github.com/Granola-Team/mina-indexer/issues/665)
> * [Problem: `config.json` output to wrong directory on startup #666](https://github.com/Granola-Team/mina-indexer/issues/666)
> 
> This PR is intended to solve #665

In that case, I am forced to approve and merge. However, I beg of you to consider my comments, above, @Isaac-DeFrain.

rust/src/bin/mina-indexer.rs
Isaac-DeFrain: will change

Isaac-DeFrain:
Closing. Will remove config.json.

Isaac-DeFrain:
Reopening. We will log the config instead of writing a file.

rust/src/bin/mina-indexer.rs
```diff
@@ -322,6 +339,8 @@ struct ServerArgsJson {
     canonical_threshold: u32,
     canonical_update_threshold: u32,
     locked_supply_csv: Option<String>,
+    web_hostname: String,
```
robinbb, line 342: Good catch!

robinbb (approved):

PR #676, opened 2024-03-26
Revert "Deprecate nix-shell use (#663)"

Isaac-DeFrain:
This reverts commit f83e8bb9e36bdb23c2d21acf9439867480ebdfc4.

Removing `nix.shell` broke support in VS Code

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,2 +1,1 @@
 use flake
-has just
--- a/Justfile
+++ b/Justfile
@@ -34,7 +34,8 @@ format:
 test: test-unit
   cd rust && ./test
 
-test-ci: lint test-unit test-regression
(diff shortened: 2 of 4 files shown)


robinbb:
@Isaac-DeFrain I don't agree that we support VSCode.

Isaac-DeFrain:
We don't now. Should we remove `./.vscode` then and modify `.gitignore`?

Isaac-DeFrain:
Closing to pursue direnv support

robinbb:
Yup, thanks for helping give direction to all these efforts, Isaac. I think it's all settled now - we don't "support" VSCode in this repo, but direnv is supported, and there is a VSCode plugin for direnv which enables Nix support.

PR #689, opened 2024-03-27
Update .gitignore

Isaac-DeFrain:


.gitignore
```diff
@@ -1,7 +1,7 @@
-/rust/target/
+target
```
robinbb, line 1: This would result in ignoring 'target' directories outside of the 'rust' directory. That means that this setting would ignore things that the dev does not know about, but would expected to be told about by 'git status'.

.gitignore
```diff
@@ -1,7 +1,7 @@
...
 result
 .gitconfig
-.direnv/
+.direnv
```
robinbb, line 4: Does this final slash not indicate that the ignored thing must be a directory?

robinbb (changes requested):
Needs well-defined problem.

.gitignore
Isaac-DeFrain: After setting up direnv, I'm getting artifacts from `just build` in `./target`

.gitignore
Isaac-DeFrain: You're probably right. In that case, shouldn't we have a slash after `result`?

jhult:
If it helps, [here](https://www.toptal.com/developers/gitignore/api/rust,rust-analyzer,macos,visualstudiocode) is how [Toptal Gignore](https://github.com/toptal/gitignore.io) recommends.

.gitignore
robinbb: That should not be. Send repro instructions?

robinbb: I guess those were repro instructions, actually. I don't get that directory. It is probably leftover from when you switched that directory from one that built in the top level to one that built in the 'rust' subdirectory.

.gitignore
robinbb: > You're probably right. In that case, shouldn't we have a slash after `result`?

Yes, and after 'target'.

robinbb:
@Isaac-DeFrain My summary of the current status is: the .gitignore as-is (before this PR) was correct in that it correctly identified an errant 'target' directory where such a directory should not be. However, it could have been made "more correct" by a suffix slash on the 'direnv' entry. So, this should not merge as-is, IMO.

.gitignore
Isaac-DeFrain: After seeing this phenomenon in multiple terminals on various branches yesterday, it appears to no longer be a problem for me. I guess, the future is bright.

Isaac-DeFrain:
No longer causing me pain

PR #691, opened 2024-03-28
Maintain same API as minaexplorer

trevorbernard:


jhult:
@trevorbernard, do only certain variables need the `#[graphql(name)` macro?

trevorbernard:
> @trevorbernard, do only certain variables need the `#[graphql(name)` macro?

It's a bug in the mina explorer API. They should all be camelCase but those exist as snake_case so I have to make them an exception

robinbb (approved):
I note absence of tests in this PR?

jhult:
> > @trevorbernard, do only certain variables need the `#[graphql(name)` macro?
> 
> It's a bug in the mina explorer API. They should all be camelCase but those exist as snake_case so I have to make them an exception

Got it. Thanks for the explanation. For the Mina Explorer GraphQL schema, I believe we are working off [this](https://github.com/Granola-Team/mina-block-explorer/blob/main/graphql/schemas/mina-explorer.graphql).

PR #702, opened 2024-03-29
Correct integration tests

### LINKED ISSUE #703: Problem: the integration/regression tests are silently failing
... due to lack of 'set -o pipefail' in the Bash script, code of the form "left | right" actually succeeds when "left" fails. :-(

robinbb:
Fixes #703 

The lack of the relatively obscure incantation 'set -o pipefail' in this script caused lines with '|' which were failing to be considered "success". This PR adds the incantation, and corrects the failing lines. Also corrected was an oversight in which 2 concurrently running tests in the same directory would use the same socket, causing both false positive tests and false negative tests, which was very confusing.

### DIFF
--- a/rust/test
+++ b/rust/test
@@ -17,6 +17,7 @@ SUMMARY_SCHEMA="$SRC"/rust/tests/json-schemas/summary.json
 # The rest of this script's logic assumes that the testing is done from within
 # this temporary directory.
 TEST_DIR="$(mktemp -d 2>/dev/null || mktemp -d -t 'mytmpdir')"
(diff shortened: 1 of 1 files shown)


robinbb:
These very difficult-to-diagnose problems speak to the use of shell languages as perhas an anti-pattern when one cares about correctness. Maybe we could try nushell to get more sanity, but still the conciseness of shell code. Or maybe simply use a simple language (Lua? Scheme? Haskell Turtle?) to express the test code. (No necessarily needing a discussion on this PR, merely musing.)

trevorbernard (approved):

PR #721, opened 2024-03-31
Properly kill processes (mina-indexers) in the regression test suite

### LINKED ISSUE #720: Problem: 'mina-indexer' processes do not die after running the regression suite
Our method of process termination is insufficient to actually kill the processes, it seems.

robinbb:
Fixes #720

### DIFF
--- a/tests/regression
+++ b/tests/regression
@@ -68,6 +68,15 @@ idxr_server() {
     echo $! > idxr_pid
 }
 
+kill_idxr() {
+    PID="$(cat ./idxr_pid)"
+    echo "Killing PID $PID."
+    kill "$PID"
+    wait "$PID" || true  # Required because the return value of 'wait' is the process's exit status.
+    echo "Killed $PID. Deleting PID file."
+    rm -f ./idxr_pid
+}
+
 idxr_server_start() {
     port=$(ephemeral_port)
     idxr_server start --web-port "$port" "$@"
@@ -79,7 +88,7 @@ idxr_server_sync() {
         --blocks-dir ./blocks \
         --ledgers-dir ./staking-ledgers \
         --log-dir ./logs \
-	"$@"
+        "$@"
 }
 idxr_server_replay() {
     port=$(ephemeral_port)
@@ -88,7 +97,7 @@ idxr_server_replay() {
         --ledgers-dir ./staking-ledgers \
         --database-dir ./database \
         --log-dir ./logs \
-	"$@"
+        "$@"
 }
 
 dl_mainnet() {
@@ -135,8 +144,7 @@ setup() {
 }
 
 teardown() {
-    kill "$(cat idxr_pid)" 2>/dev/null || true
-    rm -f ./idxr_pid
+    kill_idxr || rm -f ./idxr_pid
     rm -rf ./blocks
     rm -rf ./staking-ledgers
     rm -rf ./database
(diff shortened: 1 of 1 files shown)


tests/regression
```diff
@@ -786,10 +794,7 @@ test_sync() {
         --log-dir ./logs
     wait_for_socket
 
-    # kill running indexer and release db lock
-    kill "$(cat idxr_pid)" 2>/dev/null || true
-    rm -f idxr_pid
-    rm ./database/LOCK
```
jhult, line 792: Do we need to worry about the `./database/LOCK` anymore?

tests/regression
```diff
@@ -68,6 +68,15 @@ idxr_server() {
...
 }
 
+kill_idxr() {
+    PID="$(cat ./idxr_pid)"
+    echo "Killing PID $PID."
+    kill "$PID"
+    wait "$PID" || true  # Required because the return value of 'wait' is the process's exit status.
+    echo "Killed $PID. Deleting PID file."
+    rm -f ./idxr_pid
+}
```
jhult, line 78: I'd slightly adjust this to require each command to depend on the success of the previous. That way if for some reason it cannot find the PID, it doesn't continue trying to kill and wait for a non-existent PID.

```suggestion
kill_idxr() {
    PID="$(cat ./idxr_pid)" &&
    echo "Killing PID $PID." &&
    kill "$PID" &&
    (wait "$PID" || true) &&  # Required because the return value of 'wait' is the process's exit status.
    echo "Killed $PID. Deleting PID file."
    rm -f ./idxr_pid
}
```

tests/regression
```diff
@@ -135,8 +144,7 @@ setup() {
 }
 
 teardown() {
-    kill "$(cat idxr_pid)" 2>/dev/null || true
-    rm -f ./idxr_pid
+    kill_idxr || rm -f ./idxr_pid
```
jhult, line 147: If the above suggestion (for `kill_idxr`) is implemented, then I think this line can be:

 ```suggestion
     kill_idxr
```

File: tests/regression (hunk 1 above)
Comment (robinbb) [reply tests/regression:792]: The explicit removal of the lock file was only needed because we were silently failing to terminate the Mina Indexer process (2 lines above). This PR attempts to correct the termination, which achieves the deletion of the lock file.

File: tests/regression (hunk 2 above)
Comment (robinbb) [reply tests/regression]: This is already as you desire because 'set -e' is used at the top of the script.

File: tests/regression (hunk 3 above)
Comment (robinbb) [reply tests/regression:147]: This is an attempt to guard against a future version of 'kill_idxr' which could conceivably return an error. In that case, we still proceed to remove the PID file. You only write the line without that feature because you have knowledge of the implementation - namely, that it will never return if an error is encountered.

**Comment by robinbb** [comment]
Blocks #700

File: tests/regression (hunk 2 above)
Comment (jhult) [reply tests/regression]: @robinbb, I'm not sure `set -e` (at the top of the script) actually works in this particular context. However, I'm not going to block on this. Read [this](https://stackoverflow.com/questions/4072984/how-do-i-get-the-effect-and-usefulness-of-set-e-inside-a-shell-function) for more details.

**Review by jhult — APPROVED** [review]

File: tests/regression (hunk 2 above)
Comment (robinbb) [reply tests/regression]: @jhult The Stack Overflow answer on that page from Roman Cheplyaka is the one that is applicable in this case - the 'set -e' applies. You can test this by way of the following Bash script:

Comment (robinbb) [reply tests/regression]:
```
set -e
r() {
  echo 1
  false
  echo 2
}
r
```

This will print only '1'  and then exit.

Issue #739, opened 2024-04-03
Problem: stakes GQL endpoint doesn't filter based on public_key


trevorbernard:
The frontend requires the ability to filter based on public_key. Default to epoch 0 if none is provided

PR #740, opened 2024-04-03
(stakes) add public_key query filter

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/739

* Add ability to filter based on public_key
* default to epoch 0 if none is provided
* Update cargo.lock to fix cargo audit complaint

### DIFF
--- a/rust/src/web/graphql/stakes/mod.rs
+++ b/rust/src/web/graphql/stakes/mod.rs
@@ -57,13 +57,10 @@ impl StakeQueryRoot {
             .filter(|account| {
                 if let Some(ref query) = query {
                     if let Some(public_key) = &query.public_key {
-                        *public_key == account.pk.0
-                    } else {
-                        true
+                        return *public_key == account.pk.0;
                     }
-                } else {
-                    true
                 }
+                true
             })
             .map(|account| {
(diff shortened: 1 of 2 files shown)


rust/src/web/graphql/stakes/mod.rs
```diff
@@ -52,6 +53,17 @@ impl StakeQueryRoot {
         let mut accounts: Vec<LedgerAccountWithMeta> = staking_ledger
             .staking_ledger
             .into_values()
+            .filter(|account| {
+                if let Some(ref query) = query {
+                    if let Some(public_key) = &query.public_key {
+                        *public_key == account.pk.0
```
Isaac-DeFrain, line 59: why not `return *public_key == account.pk.0` and drop the `else`s?

Isaac-DeFrain (approved):
LGTM

rust/src/web/graphql/stakes/mod.rs
trevorbernard: if the public key isn't set you don't want to filter

Isaac-DeFrain (approved):

Issue #554, opened 2024-03-07
Problem: there is no way to query for transactions via GraphQL


trevorbernard:
Add the transaction resolver to support the call the `mina-block-explorer` uses for the frontend.

https://github.com/Granola-Team/mina-block-explorer/blob/main/graphql/queries/transactions.graphql

jhult:
Closed via #769.

PR #744, opened 2024-04-05
Enhancement: store date_time in SignedCommandWithData and add functions to better store/retrieve transactions

jhult:
This is needed as a blocker for #554.

I split this PR into multiple commits for easier review.

[move CommandStore into command/store](https://github.com/Granola-Team/mina-indexer/commit/b6230dabde23c3492db4a581fe048ff05e24b223) - This isn't strictly necessary, but seems like a good move to me. We can drop this commit if we don't want this.

### DIFF
--- a/rust/src/command/signed.rs
+++ b/rust/src/command/signed.rs
@@ -25,6 +25,7 @@ pub struct SignedCommandWithData {
     pub status: CommandStatusData,
     pub tx_hash: String,
     pub blockchain_length: u32,
+    pub date_time: u64,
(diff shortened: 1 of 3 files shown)


Isaac-DeFrain:
I strongly prefer keeping the entire impl block in one file (keep the `CommandStore` trait and the `IndexerStore` impl how/where they were), but I'm open to discussing the merits of alternatives.

Isaac-DeFrain (approved):
LGTM

Issue #698, opened 2024-03-28
Problem: GQL stakes and block endpoint are missing tests


trevorbernard:

jhult:
Implementations:

- https://github.com/Granola-Team/mina-indexer/issues/553
- https://github.com/Granola-Team/mina-indexer/issues/557

jhult:
A few possible tools:
- [hurl](https://hurl.dev/)
- [graphql-to-karate](https://github.com/wbaldoumas/graphql-to-karate)
- [k6](https://github.com/grafana/k6)
- [Step CI](https://github.com/stepci/stepci)

jhult:
Block(s) implemented via https://github.com/Granola-Team/mina-indexer/pull/792 and https://github.com/Granola-Team/mina-indexer/pull/797

jhult:
Stakes added in #800

PR #769, opened 2024-04-09
Feat: GraphQL - transactions

jhult:
Implements transactions (#554)


jhult:
I plan to add one or more tests (perhaps as part of #698 but ideally as part of this PR).

I'm moving this from _Draft_ to _Ready for review_ to get comments on the current code.

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        if let Some(__) = &self.date_time_gte {
+            matches = matches && transaction.block.date_time >= *__;
+        }
+
+        if let Some(__) = &self.date_time_lte {
+            matches = matches && transaction.block.date_time <= *__;
+        }
+
+        // TODO: implement matches for all the other optional vars
```
Isaac-DeFrain, line 203: Is it okay to leave this out for now?

Isaac-DeFrain (commented):
I'll let @trevorbernard review too. My only ask is for a regression test that queries the new endpoint

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -0,0 +1,335 @@
...
+
+        // match sort_by {
+        //     TransactionSortByInput::NonceAsc => {
+        //         transactions.sort_by(|Some(a), Some(b)| b.nonce.cmp(&a.nonce));
+        //     }
+        //     TransactionSortByInput::NonceDesc => {
+        //         transactions.sort_by(|Some(a), Some(b)| a.nonce.cmp(&b.nonce));
+        //     }
+        //     _ => {}
+        // }
```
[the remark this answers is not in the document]
jhult: It likely needs to be in there, but I had problems figuring out the proper logic and getting it to compile.

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        _query: TransactionQueryInput,
+    ) -> Result<Option<Transaction>> {
+        todo!()
+    }
+
+    pub async fn transactions(
+        &self,
+        ctx: &Context<'_>,
+        query: TransactionQueryInput,
+        limit: Option<i64>,
```
trevorbernard, line 38: I would make this a usize

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+        sort_by: TransactionSortByInput,
+    ) -> Result<Vec<Option<Transaction>>> {
+        let db = db(ctx);
+        let limit = limit.unwrap_or(100);
+        let limit_idx = limit as usize;
```
trevorbernard, line 43: see comment above

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -0,0 +1,303 @@
...
+
+            // Only add transactions that satisfy the input query
+            if query.matches(&transaction) {
+                transactions.push(Some(transaction));
+            };
+
+            // Early break if the transactions reach the query limit
+            if transactions.len() >= limit_idx {
+                break;
+            }
```
trevorbernard, line 79: The txns would have had to be in sorted order already for this early exit to be correct I think. E.g. Limit 1 and sort by NonceDesc. There would be no guarantee that the largest nonce would be returned. Unfortunately you'll need to look at an index first or scan through all the transactions

trevorbernard (changes requested):

rust/src/web/graphql/transactions/mod.rs
jhult: Excellent catch. I'll fix that.

jhult: [Fixed](https://github.com/Granola-Team/mina-indexer/pull/769/files#diff-22e46ac9e2b2d33676e8ff2a5489995088666459acf57a10bfa88b6ea6c5dac1R77) by removing `if` statement and using `truncate`.

trevorbernard (approved):

PR #801, opened 2024-04-23
Snarks gql endpoint

trevorbernard:


tests/hurl/blocks.hurl
```diff
@@ -47,4 +47,4 @@ POST {{url}}
 ```
 HTTP 200
 [Asserts]
-jsonpath "$.data.blocks" count == 2
+jsonpath "$.data.blocks" count == 3
```
jhult, line 50: What are these changes to the tests for blocks and fee transfers?

Isaac-DeFrain: There was something wrong with the global slot prefix for blocks and it was over-writing data

trevorbernard: This wasn't that issue. We are ingesting 20 more blocks with the hurl tests and the results changed.

PR #804, opened 2024-04-24
Refactor GQL modules

Isaac-DeFrain:


rust/src/web/rest/blocks.rs
```diff
@@ -30,22 +30,21 @@ pub async fn get_blocks(
...
                 parent_state_hash = block.previous_state_hash();
                 best_chain.push(block);
             } else {
                 // No parent
                 break;
             }
-            counter += 1;
+            if best_chain.len() == limit as usize {
+                break;
+            }
```
jhult, line 45: Could this ever get stuck and never hit the break?

rust/src/web/mod.rs
```diff
@@ -49,7 +49,7 @@ pub async fn start_web_server<A: net::ToSocketAddrs, P: AsRef<Path>>(
             .service(
                 web::resource(ENDPOINT_GRAPHQL)
                     .guard(guard::Get())
-                    .to(index_graphiql),
+                    .to(indexer_graphiql),
```
jhult, line 52: This is fine as either `index_graphiql` or `indexer_graphiql`. I think the original intention as `index` was meaning it sits at the index of the host (e.g. `http://host/`)

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -96,16 +98,12 @@ impl TransactionsQueryRoot {
...
-        None => false,
-    };
-
-    Transaction::from_cmd(cmd, block_date_time, &block_state_hash, canonical)
+    Transaction::from_cmd(
+        cmd,
+        block_date_time,
+        &block_state_hash,
+        get_block_canonicity(db, &block_state_hash.0),
+    )
```
jhult, line 106: Just inlining the call to `get_block_canonicity`; makes sense.

rust/src/web/graphql/transactions/mod.rs
```diff
@@ -31,14 +30,25 @@ impl TransactionsQueryRoot {
...
+        let db = db(ctx);
         if let Some(hash) = query.hash {
             if signed::is_valid_tx_hash(&hash) {
-                let db = db(ctx);
                 return Ok(db
                     .get_command_by_hash(&hash)?
                     .map(|cmd| txn_from_hash(cmd, db)));
             }
+        } else {
+            // no query filter => return the most recent transaction
```
jhult, line 41: This seems like functionality that should be documented (more so than in just a code comment).

rust/src/store.rs
```diff
@@ -204,22 +210,13 @@ impl BlockStore for IndexerStore {
...
-            let key = global_slot_block_key(block);
-            let value = block.state_hash.as_bytes();
-            let blocks_global_slot_idx_cf = self.blocks_global_slot_idx_cf();
-            self.database
-                .put_cf(&blocks_global_slot_idx_cf, key, value)?;
-        }
+        self.database.put_cf(
+            self.blocks_global_slot_idx_cf(),
+            global_slot_block_key(block),
+            b"",
```
trevorbernard, line 217: This is likely a bug. We should be inserting the `state_hash`

rust/src/web/graphql/blocks/mod.rs
```diff
@@ -32,49 +22,56 @@ impl BlocksQueryRoot {
...
+            })?);
+        }
+
         // Use constant time access if we have state hash
         if let Some(state_hash) = query.clone().and_then(|input| input.state_hash) {
-            let state_hash = &BlockHash::from(state_hash);
-            let pcb = match db.get_block(state_hash)? {
+            if !block::is_valid_state_hash(&state_hash) {
+                return Ok(None);
+            }
```
trevorbernard, line 39: I don't think this check is necessary. It's extra work for the same result.

rust/src/web/rest/blocks.rs
Isaac-DeFrain: It's semantically equivalent to what was there, just without the `counter` variable. Each time we loop, it either pushes a block or breaks if it has no parent

rust/src/web/graphql/transactions/mod.rs
Isaac-DeFrain: There were several versions of `get_block_canonicity` in the GQL modules. Those all got refactored to this common version

rust/src/web/graphql/transactions/mod.rs
Isaac-DeFrain: Agreed. Is this the expected behavior for query `transactions {...}`?

rust/src/store.rs
Isaac-DeFrain: This is a feature, we don't need to store the state hash in both the key and the value. The key has the form: `{4 global slot BE bytes}{state hash bytes}`.

rust/src/web/graphql/blocks/mod.rs
Isaac-DeFrain: You're right, but it's less work for an invalid state hash

rust/src/web/graphql/transactions/mod.rs
Isaac-DeFrain: Or should we return the default limit number of transactions?

jhult: @Isaac-DeFrain, I would think return the default limit number of transactions. I made an issue: https://github.com/Granola-Team/mina-indexer/issues/805

PR #831, opened 2024-05-01
Add Berkeley block parser

### LINKED ISSUE #327: Problem: We don't have a `berkeley` precomputed block parser
Use the `testworld` precomputed blocks and create a a serde serialize and deserialize functions

Isaac-DeFrain:
Fixes #327


trevorbernard (approved):
I went through it as thoroughly as one could.

Issue #836, opened 2024-05-06
Problem: nextStakes and stakes GQL endpoint has the same schema (mostly) but duplicate objects


trevorbernard:
Reuse what you can from the Stakes GQL endpoint

PR #837, opened 2024-05-06
(gql) reuse stakes objects where applicable in the next stakes API

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/836

* reuse stakes GQL objects where applicable in the next stakes GQL resolver


jhult (approved):
Nice!

PR #841, opened 2024-05-07
feat: use network from db

### LINKED ISSUE #655: Support querying multiple networks
We should expand the cli query args to have an optional flag `--network` which defaults to `"mainnet"`

jhult:
Fixes #655


Isaac-DeFrain (approved):
LGTM

Issue #852, opened 2024-05-09
Problem: transaction(s) GQL endpoints returning incorrect values for amount and fee


trevorbernard:
We're return back a mina float where the value needs to be in nanomina

PR #853, opened 2024-05-09
Complete transaction to frontend integration

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/848
Fixes: https://github.com/Granola-Team/mina-indexer/issues/852

* Add efficient filter by block height
* Use nanomina for transaction amount and fee


rust/src/web/graphql/transactions/mod.rs
```diff
@@ -59,6 +60,20 @@ impl TransactionsQueryRoot {
             }
         };
 
+        // block height query
+        if let Some(block_height) = query.block_height {
+            let mut transactions: Vec<TransactionWithBlock> = db
+                .get_blocks_at_height(block_height as u32)?
```
jhult, line 66: Should we (somehow) support providing a limit when querying the database?

I suppose we can't do that due to sorting.

tests/hurl/transactions.hurl
```diff
@@ -40,3 +40,49 @@ jsonpath "$.data.transactions[49].memo" == "E4YM2vTHhWEg66xpj52JErHUBU4pZ1yageL4
...
+  "query": {
+    "canonical": true,
+    "blockHeight": 11
+  }
+}
+```
+
+HTTP 200
+[Asserts]
+jsonpath "$.data.transactions" count == 2
```
jhult, line 81:
```suggestion
duration < 1000
jsonpath "$.data.transactions" count == 2
```

jhult (approved):
LGTM. Nicely done!

rust/src/web/graphql/transactions/mod.rs
trevorbernard: This won't ever be an expensive call. I've only ever seen 2-3 blocks per block height and there are a max of 128 txns per block.

jhult (approved):

PR #865, opened 2024-05-10
(gql) use public key index when transactions filter include from or to

trevorbernard:


jhult:
To confirm, we aren't currently supporting querying/filtering against multiple things such as to/from and block height, correct? If they send in both, it will use whichever we have defined first in our (internal) implementation.

trevorbernard:
> To confirm, we aren't currently supporting querying/filtering against multiple things such as to/from and block height, correct? If they send in both, it will use whichever we have defined first in our (internal) implementation.

@jhult This was a draft to send to @Isaac-DeFrain . It was to illustrate that the index we have available isn't fast enough. The outcome is that we need two indicies for `from` and `to`. It will look something like this:

```
fromCF: {send_pk}{global_slot}{txn_hash} -> amount
toCF:   {recv_pk}{global_slot}{txn_hash} -> amount
```

So this allows us to use a prefix iterator on either the `sender` or `receiver` and have the txns sorted by global slot height.

Issue #211, opened 2023-10-06
Problem: We rely on O(1)'s precomputed blocks bucket 


trevorbernard:
Run a Mina node and log the precomputed blocks into a bucket we control. E.g. S3 Bucket

Isaac-DeFrain:
Just need to resolve merge conflicts https://github.com/MinaProtocol/mina/pull/13236. However, this solution will only allow us to get blocks in the transition frontier. Hosting a bucket of all precomputed blocks is another story altogether.

robinbb:
@Isaac-DeFrain My interpretation of this issue is that it is about the hosting of a precomputed blocks bucket. I'll take it.

robinbb:
`https://granola-mina-blocks.se-sto-1.linodeobjects.com` is now where we host our (corrected) block files. Also, we do not use gsutil to download blocks in our regression suite.

A final step is to create a Mina node that uploads blocks to the bucket in an automated fashion.

jhult:
Related to https://github.com/Granola-Team/mina-indexer/issues/609

PR #886, opened 2024-05-13
Add example scripts to help with onboarding

n1tranquilla:


scripts/download-blocks.sh
```diff
@@ -0,0 +1,17 @@
...
+MAX_LENGTH=200
+
+OUT_DIR=.
+
+for LENGTH in `seq $MIN_LENGTH $MAX_LENGTH`
+do
+    echo "gs://mina_network_block_data/mainnet-${LENGTH}-*.json" >> "$BLOCKS_DIR"/block_urls
+done 2>/dev/null
+
+gsutil -m cp -n -I "$BLOCKS_DIR" < "$BLOCKS_DIR"/block_urls
```
jhult, line 17: I think we can/should get rid of `gsutil` and download from our own bucket

See https://github.com/Granola-Team/mina-indexer/issues/211#issuecomment-2010888237

jhult (dismissed):
Also requesting to squash commits.

scripts/nuke-database.sh
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env bash
+
+rm -rf ~/.mina-indexer/database
```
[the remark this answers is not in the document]
n1tranquilla: Yeah, I mean that would make sense. Although the intent is to use at your own risk.

scripts/download-blocks.sh
n1tranquilla: That would be cool. Would these scripts be targeted to Granolas or to the broader community as well?

jhult: I think targeting the broader community would be a good idea.

scripts/download-blocks.sh
```diff
@@ -0,0 +1,17 @@
+#!/bin/sh
+set -ex
+
+BLOCKS_DIR="./blocks"
+touch "$BLOCKS_DIR"/block_urls
+
+MIN_LENGTH=1
```
trevorbernard, line 7: This will fail. Blocks start at 2

scripts/download-blocks.sh
trevorbernard: That's a different issue and shouldn't block this PR

scripts/nuke-database.sh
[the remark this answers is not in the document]
Isaac-DeFrain: I will approve and we can add this later if needed.

Isaac-DeFrain (approved):
LGTM

trevorbernard (approved):

Issue #862, opened 2024-05-10
Internal Commands GQL endpoint should support querying by `recipient`


trevorbernard:
```json
{
    "data": null,
    "errors": [
        {
            "message": "Invalid value for argument \"query\", unknown field \"recipient\" of type \"FeetransferQueryInput\"",
            "locations": [
                {
                    "line": 6,
                    "column": 49
                }
            ]
        }
    ]
}
```

PR #888, opened 2024-05-13
(gql) feetransfers filter by recipient

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/862

* Filter by recipient in feetransfers

### DIFF
--- a/tests/hurl/feetransfers.hurl
+++ b/tests/hurl/feetransfers.hurl
@@ -29,3 +29,20 @@ jsonpath "$.data.feetransfers[0].blockHeight" == 120
 jsonpath "$.data.feetransfers[0].canonical" == true
 jsonpath "$.data.feetransfers[0].dateTime" == "2021-03-17T08:39:00.000Z"
 duration < 1000
+
+POST {{url}}
+```graphql
+{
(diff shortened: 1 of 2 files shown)


tests/hurl/feetransfers.hurl
```diff
@@ -29,3 +29,20 @@ jsonpath "$.data.feetransfers[0].blockHeight" == 120
...
+    fee
+    recipient
+    type
+    dateTime
+  }
+}
+```
+HTTP 200
+[Asserts]
+jsonpath "$.data.feetransfers" count == 1
```
jhult, line 46: Will there only ever be 1 or do we need to also query by epoch?

jhult (approved):

tests/hurl/feetransfers.hurl
trevorbernard: This will only every be 1 in the first 120 blocks. Also, filtering by epoch doesn't make sense in this cause as it's not a staking ledger. I could narrow it down by including a staking ledger if that's what you want

jhult (approved):

PR #905, opened 2024-05-17
Feat/db sorting for gql

### LINKED ISSUE #863: Querying the transactions GQL endpoint using `from` or `to` attribute should use an index
Create two new indices for the `sender` and `receiver` portion of the transaction. This will look something like the following:

```
fromCF: {send_pk}{global_slot}{txn_hash} -> amount
toCF:   {recv_pk}{global_slot}{txn_hash} -> amount
```

So this allows us to use a prefix iterator on either the sender or receiver and have the transactions sorted by global slot height.

### LINKED ISSUE #873: Querying for all user commands for a public key is too slow
We currently have the following index: `get_commands_for_public_key` but on large public keys there may be 10's of thousands but by default, there is no global slot sorting (aka by wall clock). So for us to achieve correctness on sort ordering, we must process all transactions, then sort and limit; this is too slow.

Proposed solution:

Create two new indices for the `sender` and `receiver` portion of the transaction. This will look something like the following:

```
fromCF: {send_pk}{global_slot}{txn_hash} -> amount
toCF:   {recv_pk}{global_slot}{txn_hash} -> amount
```

So this allows us to use a prefix iterator on either the sender or receiver and have the transactions sorted by global slot height.

### LINKED ISSUE #894: Cannot sort accounts in db by `balance` for GQL queries

Isaac-DeFrain:
There are a lot of changes in this PR all related to storing/sorting data.

- fixes #894
- fixes #863
- fixes #873
- removes unnecessary best ledger clone in `summary` REST endpoint

### DIFF
--- a/rust/src/block/mod.rs
+++ b/rust/src/block/mod.rs
@@ -45,6 +45,9 @@ pub struct BlockWithoutHeight {
 pub struct BlockHash(pub String);
 
 impl BlockHash {
+    pub const LEN: usize = 52;
+    pub const PREFIX: &'static str = "3N";
+
     pub fn from_bytes(bytes: &[u8]) -> anyhow::Result<Self> {
         let res = String::from_utf8(bytes.to_vec())?;
         if is_valid_state_hash(&res) {
@@ -279,7 +282,7 @@ impl std::fmt::Display for BlockHash {
 pub fn get_state_hash(file_name: &OsStr) -> Option<String> {
     let last_part = file_name.to_str()?.split('-').last()?.to_string();
     let state_hash = last_part.split('.').next()?;
-    if state_hash.starts_with("3N") {
+    if state_hash.starts_with(BlockHash::PREFIX) {
         return Some(state_hash.to_string());
     }
     None
@@ -297,7 +300,7 @@ pub fn get_blockchain_length(file_name: &OsStr) -> Option<u32> {
 }
 
 pub fn is_valid_state_hash(input: &str) -> bool {
-    input.starts_with("3N") && input.len() == 52
+    input.starts_with(BlockHash::PREFIX) && input.len() == BlockHash::LEN
 }
 
 pub fn is_valid_file_name(path: &Path, hash_validator: &dyn Fn(&str) -> bool) -> bool {
--- a/rust/src/block/mod.rs
+++ b/rust/src/block/mod.rs
@@ -18,7 +18,7 @@ use crate::{
         version_bytes,
     },
 };
-use anyhow::bail;
+use anyhow::{anyhow, bail};
 use serde::{Deserialize, Serialize};
 use std::{ffi::OsStr, path::Path};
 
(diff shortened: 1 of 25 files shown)


rust/src/command/mod.rs
```diff
@@ -142,7 +142,11 @@ pub trait UserCommandWithStatusT {
...
     fn sender(&self) -> PublicKey;
 
+    fn receiver(&self) -> PublicKey;
+
     fn nonce(&self) -> u32;
+
+    fn amount(&self) -> u64;
 }
 
 impl UserCommandWithStatusT for UserCommandWithStatus {
```
trevorbernard, line 152: UserCommandWithStatusT?

rust/src/ledger/public_key.rs
```diff
@@ -9,13 +9,30 @@ use serde::{Deserialize, Serialize};
 pub struct PublicKey(pub String);
 
 impl PublicKey {
+    pub const LEN: usize = 55;
+    pub const PREFIX: &'static str = "B62q";
```
trevorbernard, line 13: This seems suspicious. Do we know that this will always be the prefix?

rust/src/block/mod.rs
```diff
@@ -45,15 +45,28 @@ pub struct BlockWithoutHeight {
 pub struct BlockHash(pub String);
 
 impl BlockHash {
-    pub fn from_bytes(bytes: [u8; 32]) -> Self {
-        let block_hash = unsafe { String::from_utf8_unchecked(Vec::from(bytes)) };
-        Self(block_hash)
+    pub const LEN: usize = 52;
+    pub const PREFIX: &'static str = "3N";
```
trevorbernard, line 49: You don't need to `&'static` prefix if it's const

rust/src/block/store.rs
```diff
@@ -44,6 +47,19 @@ pub trait BlockStore {
     /// Add a block at the given global slot since genesis
     fn add_block_at_slot(&self, state_hash: &BlockHash, slot: u32) -> anyhow::Result<()>;
 
+    /// TODO
```
trevorbernard, line 50: Is this valid?

rust/src/block/store.rs
```diff
@@ -62,4 +78,28 @@ pub trait BlockStore {
...
+    fn set_block_balance_updates(
+        &self,
+        state_hash: &BlockHash,
+        balance_updates: Vec<PaymentDiff>,
+    ) -> anyhow::Result<()>;
+
+    /// TODO
+    fn get_block_balance_updates(&self, state_hash: &BlockHash)
+        -> anyhow::Result<Vec<PaymentDiff>>;
```
trevorbernard, line 101: What's with all the TODOs?

trevorbernard:
There is way too much going on in this PR IMHO but good work

rust/src/command/mod.rs
Isaac-DeFrain: Trait for `UserCommand` functions. It was added with v2 blocks.

rust/src/ledger/public_key.rs
Isaac-DeFrain: We do. We've been using them for a long time now. Just wanted to make them a little handier.

rust/src/block/store.rs
Isaac-DeFrain: Draft. Will leave comments

rust/src/block/store.rs
Isaac-DeFrain: 💯

rust/src/block/store.rs
Isaac-DeFrain: Just wait until you hear what that `common_ancestor_account_balance_updates` is for!

rust/src/block/mod.rs
Isaac-DeFrain: image

jhult (approved):
Well done, sir!

PR #909, opened 2024-05-18
Problem: Nix image tag uses 9 digits (1 too many)

robinbb:
Solution: Use only 8 digits in tag, consistent with output of `git rev-parse --short`.


jhult (approved):
LGTM.

I will test merging before CI tests are complete.

PR #918, opened 2024-05-22
Problem: Robin can't count to 8

robinbb:
Corrects a failure in the "create OCI image" step due to the image label having 1 digit too few.


jhult (approved):
😂

Issue #928, opened 2024-05-23
Problem: blocks GQL endpoint panics on a simple query


trevorbernard:
Steps to recreate the panic
```graphl
{
  blocks(limit: 10) {
    stateHash
  }
}

```

PR #929, opened 2024-05-23
Add regression test for blocks panic

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/928

* use `map_or` rather than `unwrap` on block height and global slot bounded queries


Isaac-DeFrain (approved):
LGTM

Issue #933, opened 2024-05-23
Problem: incorrect GQL schema for blocks global slot since genesis


trevorbernard:
This is the correct schema 
```graphql
{
  blocks(limit: 1) {
    protocolState {
      consensusState {
        slotSinceGenesis
      }
    }
  }
}
```

PR #937, opened 2024-05-23
Filter blocks global slot

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/933

* Add correct slotSinceGenesis schema

### DIFF
--- a/rust/src/web/graphql/blocks/mod.rs
+++ b/rust/src/web/graphql/blocks/mod.rs
@@ -131,11 +131,14 @@ impl BlocksQueryRoot {
         }
 
         // global slot query
-        if let Some(global_slot_since_genesis) =
-            query.as_ref().and_then(|q| q.global_slot_since_genesis)
+        if let Some(global_slot_since_genesis) = query
+            .as_ref()
+            .and_then(|q| q.protocol_state.as_ref())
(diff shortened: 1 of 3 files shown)


rust/src/web/graphql/blocks/mod.rs
```diff
@@ -131,11 +131,14 @@ impl BlocksQueryRoot {
         }
 
         // global slot query
-        if let Some(global_slot_since_genesis) =
-            query.as_ref().and_then(|q| q.global_slot_since_genesis)
+        if let Some(global_slot_since_genesis) = query
+            .as_ref()
+            .and_then(|q| q.protocol_state.as_ref())
```
Isaac-DeFrain, line 136: Should this be `.or()`? Otherwise the query must provide them all

Isaac-DeFrain (changes requested):

rust/src/web/graphql/blocks/mod.rs
robinbb: @Isaac-DeFrain If you are certain, then you can push a commit with a change. Otherwise, you'll have to merge this PR unless it is malformed.

Isaac-DeFrain (approved):

PR #940, opened 2024-05-23
Adjust for erratic length returned by git rev-parse

robinbb:


ops/build-oci-image
```diff
@@ -2,10 +2,11 @@
...
+IMAGE=mina-indexer:"$REV"
 docker --version
 nix build .#dockerImage
 docker load < ./result
-docker run --rm -it \
-  mina-indexer:"$(git rev-parse --short HEAD)" \
+docker run --rm -it "$IMAGE" \
   mina-indexer server start --help
-docker image rm mina-indexer:"$(git rev-parse --short HEAD)"
+docker image rm "$IMAGE"
```
jhult, line 12: Eventually, we will have a `push` (to a container repo) before the `image rm`, correct?

jhult (approved):
LGTM

ops/build-oci-image
robinbb: Yes, eventually. But, for now, we merely build the image and check to see if the Indexer binary therein works.

Issue #987, opened 2024-05-29
Problem: feetransfers GQL return incorrect results when querying by `state_hash` and other attributes


trevorbernard:
We optimized when filtering by `state_hash` case but forgot to match on the other predicates

PR #988, opened 2024-05-29
(gql) fix incorrect filtering when state_hash and other attributes are set

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/987

* Fix state_hash filtering optimization
* Improve performance of the matching in the conjunction case by early exiting


Isaac-DeFrain (approved):
LGTM

Issue #1003, opened 2024-05-31
Problem: filtering by `blockHeight_lte` for blocks GQL doesn't early exit


trevorbernard:
This becomes apparent when dealing with a large amount of blocks.

PR #1005, opened 2024-05-31
(gql) fix blockHeight_lte filtering not earlying exiting for blocks

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1003

* Fix early exit (performance issue) in the blocks GQL endpoint when the `blockHeight_lte` is set


Isaac-DeFrain (approved):
LGTM

PR #1020, opened 2024-05-31
Use rclone for bucket downloads

robinbb:
Was having huge problems with s3cmd. Extreme slowness. Even opened a ticket with Linode, but the problem was s3cmd. I have replaced it with the awesome 'rclone'.


robinbb:
Tested this on the builder machine to install more blocks for the tier2 testing, FYI.

PR #1022, opened 2024-06-02
fix(memo): don't panic on memo decoding

jhult:
Use default if we fail to decode.


robinbb:
Fixed elsewhere. Closing.

PR #1030, opened 2024-06-03
Follow reference implementation for decoding memos

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1029

* Drop tag and length before converting to UTF8 string
* Use length byte to set the size of the stringified memo


Isaac-DeFrain (approved):
LGTM

Issue #1027, opened 2024-06-03
Query by SNARKS GQL performance regression


trevorbernard:
The following should be instantaneous but now takes 10s of seconds

```graphql
query SnarksQuery(
  $sort_by: SnarkSortByInput!
  $limit: Int = 10
  $query: SnarkQueryInput!
) {
  snarks(sortBy: $sort_by, limit: $limit, query: $query) {
    blockHeight
    dateTime
    prover
    canonical
    block {
      stateHash
    }
    fee
  }
}

variables {
  "sort_by": "BLOCKHEIGHT_DESC",
  "limit": 10,
  "query": {
    "canonical": true
  }
}
```

PR #1031, opened 2024-06-03
(gql) fix SNARKs GQL performance regression

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1027

* in the general case, we didn't early exit


Isaac-DeFrain (approved):
LGTM

Issue #1054, opened 2024-06-05
Problem: server panick in block_store_impl


robinbb:
Found the following in the server logs:
```
thread 'actix-server worker 0' panicked at src/store/block_store_impl.rs:109:57:   
called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum PrecomputedBlock", line: 0, column: 0)
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace  
thread 'actix-server worker 6' panicked at src/store/block_store_impl.rs:109:57:
called `Result::unwrap()` on an `Err` value: Error("data did not match any variant of untagged enum PrecomputedBlock", line: 0, column: 0)
```

This came at the very end of the server log, and the PID for the server still exists. It seems still to be functioning, so this panick did not kill the server.

Isaac-DeFrain:
I may have introduced this bug

robinbb:
@Isaac-DeFrain Fair to leave this assigned to you, and to mark it "In Progress"?

Isaac-DeFrain:
I added context to the error so it will at least print the bytes it was trying to deserialize

robinbb:
Okay, then if I understand correctly, you are not doing anything more about it, you are blocked on not seeing another crash (which should log more data to help to diagnose). So, I'll mark this 'blocked'.

robinbb:
Actually, if I'm correct about the above, then I should take this issue and periodically check prod's logs for panics.

robinbb:
@Isaac-DeFrain Do you think that this bug might have occurred because data was being added into one of the directories being watched, and the data was not "complete" yet, meaning that the file was not yet fully copied at the time that we tried reading it?

Isaac-DeFrain:
That's certainly a possibility. Maybe @trevorbernard has some insight?

robinbb:
Haven't seen this again. Closing.

PR #1057, opened 2024-06-05
Add context to get_block

Isaac-DeFrain:
- adds context to diagnose #1054


Isaac-DeFrain:
I missed this in the last PR

Issue #1060, opened 2024-06-05
Blocks GQL and Summary REST API slot value incorrect


trevorbernard:
It's returning the global slot

Issue #1058, opened 2024-06-05
Problem: %ge complete in staking page displays more than 100%


robinbb:
![Image](https://github.com/Granola-Team/mina-block-explorer/assets/550475/260978e0-b551-431f-b304-2e7fa16d1d36)

trevorbernard:
I'll look into it

trevorbernard:
Created the following issue: https://github.com/Granola-Team/mina-indexer/issues/1060

PR #1061, opened 2024-06-05
Correctly calculate the slot for blocks and summary API

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1060
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1058

* Correctly return slot value for the block


robinbb:
@trevorbernard Very nice!

PR #1081, opened 2024-06-06
Add `GIT_COMMIT_HASH` to .envrc

Isaac-DeFrain:


.envrc
```diff
@@ -1,2 +1,3 @@
 use flake
 has just
+export GIT_COMMIT_HASH=$(git rev-parse --short=8 HEAD)
```
robinbb, line 3: This will not update when you commit, but okay.

trevorbernard: It's more to fix our build tools. Another alternative is to have just reload the direnv file on build

trevorbernard: just setting it to `export GIT_COMMIT_HASH=` is sufficient to make our tools work.

robinbb (approved):

PR #1091, opened 2024-06-07
Fix transaction storage

### LINKED ISSUE #1072: Fix transaction storage
_Problem_: We are only storing transactions at their hash, but we need to keep separate versions for every block they appear in.

_Solution_: Store transactions at `{txn_hash}{state_hash}` instead

Isaac-DeFrain:
- resolves #1072


trevorbernard (approved):

trevorbernard:
There are a couple of small issues that I noticed but we'll create separate issues for them

robinbb:
So many improvements!

Issue #1106, opened 2024-06-11
Txns duplicated for each block


Isaac-DeFrain:
![Image](https://github.com/Granola-Team/mina-indexer/assets/42081488/a0c4be7e-eecd-44a6-9916-ace77de999e5)

robinbb:
The 'canonical' filter in the upper right should mean that only the canonical user commands are shown.

robinbb:
Moving to Integrate milestone because this issue blocks another in that milestone.

robinbb:
@trevorbernard Isaac says this one is up for grabs.

Issue #1079, opened 2024-06-06
Problem: transaction GQL query input inconsistent with other endpoints


trevorbernard:
Make query, limit, and sortBy optional QueryInput attributes to be consistent with ME and other endpoints

PR #1116, opened 2024-06-12
Correctly query txns when filtering by `from` or `to`

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1106
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1079

* Correctly query txns when filtering by `from` or `to`
* Make query and sort_by optional


Isaac-DeFrain (approved):
LGTM

PR #1159, opened 2024-06-15
Faster dev build

robinbb:
- Removes 30 dependency packages in Cargo.lock over and above the 12 dependency packages removed in the target branch.
- Makes the dev build faster
- Introduces 'just check' for quick checks of buildability


n1tranquilla:
Approve

Isaac-DeFrain (approved):
LGTM

PR #1162, opened 2024-06-15
Build timings

robinbb:
Adds timings to build steps so that we can know how long they are taking, so that we can improve them.

Additionally changes our build so that it uses the Nix version of the indexer to run the Tier-2 testing. This is so that we test that version - which is the one that is intended to be eventually distributed on an OCI image.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -32,7 +32,7 @@ prereqs:
 
 build:
   echo "--- Performing build"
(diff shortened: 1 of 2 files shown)


robinbb:
I have tested this on the machine that runs the Tier-2 tests, FYI, to reduce the likelihood that it will break the Tier-2 pipeline.

PR #1164, opened 2024-06-15
Remove unused cargo dependencies

robinbb:
Removes 12 unused package dependencies from Cargo.lock.

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -30,42 +30,12 @@ prereqs:
   hurl --version
   shellcheck --version
 
-build:
-  echo "--- Performing build"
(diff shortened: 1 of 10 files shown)


robinbb:
Makes a significant dent in the reduction of dependencies. Also removes the unnecessary compression on incoming/outgoing queries. (The latter is okay because we use a tunnel to Cloudflare to deploy, and that tunnel does the encryption and compression for us.)

n1tranquilla:
Approved.

Isaac-DeFrain (approved):
Very nice!

PR #1167, opened 2024-06-16
Fixes to make CI tier-2 work

robinbb:


robinbb:
Required to fix tier-2 testing (which is currently red).

robinbb:
Docker image is only 37 MB!

Issue #1200, opened 2024-06-22
Problem: filtering GQL SNARKs by prover produces (potentially) incorrect results


trevorbernard:
Iterate over all the blocks at global slot vs only grabbing the first.

PR #1202, opened 2024-06-22
(gql) Iterate all blocks at slot height in prover snarks GQL query

trevorbernard:
Fixes: https://github.com/Granola-Team/mina-indexer/issues/1200

* Iterate all blocks at global slot height in prover SNARKs GQL query

### DIFF
--- a/rust/src/web/graphql/snarks/mod.rs
+++ b/rust/src/web/graphql/snarks/mod.rs
@@ -228,36 +228,36 @@ impl SnarkQueryRoot {
             };
 
             let mut snarks = Vec::new();
-            for (key, snark) in db.snark_prover_iterator(mode).flatten() {
+            'outer: for (key, snark) in db.snark_prover_iterator(mode).flatten() {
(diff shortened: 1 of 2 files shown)


rust/src/web/graphql/snarks/mod.rs
```diff
@@ -228,36 +228,36 @@ impl SnarkQueryRoot {
...
+                            snark,
+                            state_hash,
+                            db.get_snarks_epoch_count(None).expect("epoch snarks count"),
+                            db.get_snarks_total_count().expect("total snarks count"),
+                        )
+                            .into(),
+                    };
+                    if query.as_ref().map_or(true, |q| q.matches(&sw)) {
+                        snarks.push(sw);
+                        if snarks.len() == limit {
```
robinbb, line 258: In situations like this, consider instead checking `snarks.len() >= limit` out of paranoia that some future code change will result in overfilling the `snarks` by a little.

robinbb (approved):

PR #1231, opened 2024-07-01
feat: write running PID to file; prevent new instance if PID is running 

### LINKED ISSUE #1169: Problem: can't find process using a given indexer database
Given a particular mina-indexer database directory, a sysadmin would like to know which process is using the database. A mina-indexer process might also like to know if the database directory is in use by another running process. This is because it is impossible (unsupported) for more than one Indexer to use a given database directory at a time. Without this, it is impossible to safely re-use a database directory that exists when starting a mina-indexer server. Since it takes hours to re-create a database directory from logs, this is a major problem.

jhult:
Fixes #1169

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,10 +2,8 @@
 /result
 /.gitconfig
 
-# database and snapshots
+# local database
 database/
-*.gz
-PID
 
 # direnv creates this directory as a cache
 .direnv
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 # database and snapshots
 database/
 *.gz
+PID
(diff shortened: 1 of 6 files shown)


robinbb:
I have removed the regression test. It would have been very difficult to get right because the entire premise of the .bash file is that commands aren't allowed to fail, so checking for failure (return code) from a subcommand is very difficult in that file. Soon, the ops code will be the (regression) test of that feature, and its absence will be noticeable. In other words, there will be a test for it. (But not a unit test. Feel free to add a unit test.)

robinbb (approved):

Issue #1238, opened 2024-07-02
pwait is missing on macOS


jhult:
https://stackoverflow.com/questions/1058047/wait-for-a-process-to-finish

jhult:
I tried several solutions from the StackOverflow question. However, I was not able to make those work.

I think rewriting the waiting for process in Ruby may be the best solution. Assigning back to @robinbb.

trevorbernard:
As Robin mentioned, it's probably best to implement this in Ruby instead of "bash-and-friends" But this is known to work with MacOS - https://github.com/freebsd/freebsd-src/tree/main/bin/pwait

robinbb:
Now, pwait does not appear in our codebase.

PR #1239, opened 2024-07-02
feat: implement pwait function

jhult:
Fixes Granola-Team/mina-indexer#1238


trevorbernard:
This is a fix to support mac right?

robinbb:
Tough to make work. This method may be doomed to have problems. I propose this be closed, and the "wait for process" functionality coded in Ruby instead of Bash-and-friends.

PR #1242, opened 2024-07-02
fix(snapshot): include INDEXER_VERSION file

### LINKED ISSUE #1241: Snapshot does not include the `INDEXER_VERSION` file
Solution: include it

jhult:
Fixes #1241

Also fixes issue with database not waiting to be shutdown


robinbb:
Very nice!

Issue #1243, opened 2024-07-03
Problem: database is not compressed (flushed) except by a snapshot


robinbb:
Only a snapshot operation seems to force a database compression/flushing. Before a flush, the prod database occupies 360 GB; after a flush, it occupies 50 GB. The extra space usage is unnecessary and risks overfilling the disk, causing Indexer failure. It is also not possible to create another database (of a different version, say) on the same machine due to the disk being overfull (unnecessarily).

A probable solution is to executed a flush operation immediately after initial ingestion completes.

PR #1245, opened 2024-07-03
feat: flush/compress database after ingestion

jhult:
Fixes Granola-Team/mina-indexer#1243

### DIFF
--- a/rust/src/server.rs
+++ b/rust/src/server.rs
@@ -18,6 +18,7 @@ use crate::{
 };
 use log::{debug, error, info, trace};
 use notify::{Config, Event, EventKind, RecommendedWatcher, RecursiveMode, Watcher};
+use speedb::checkpoint::Checkpoint;
 use std::{
     collections::HashSet,
     fs,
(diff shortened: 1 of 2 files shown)


rust/src/server.rs
```diff
@@ -259,6 +261,15 @@ fn initialize(
     );
     state.phase = IndexerPhase::Watching;
 
+    // flush/compress database
+    let random_number = rand::thread_rng().gen_range(1..=100);
+    let temp_checkpoint_dir = std::env::temp_dir().join(format!("snapshot_{random_number}"));
```
robinbb, line 266: Need to make sure that the temporary directory used is on the same volume as the database. Best to create an "adjacent" file in the filesystem, there. Otherwise, the temporary directory will not likely have enough space.

robinbb:
@jhult Please make a change to the database version with this PR - maybe a patch-level increase? Otherwise, this functionality will not be made use of because the prior database will simply be re-used.

rust/src/server.rs
```diff
@@ -259,6 +261,15 @@ fn initialize(
     );
     state.phase = IndexerPhase::Watching;
 
+    // flush/compress database
+    let random_number = rand::thread_rng().gen_range(1..=100);
```
robinbb, line 265: Suggest avoiding the dependency on 'rand' by simply choosing a suffix ('tmp-checkpoint') and appending it. It is not likely to last a long time on the filesystem, so little chance of conflict with another temporary snapshot.

rust/src/server.rs
trevorbernard: Also, if it's not on the same device, the checkpoint will copy the files over rather than create hard-links

robinbb (approved):
Good stuff. Let's merge this and see how it goes.

PR #1249, opened 2024-07-05
chore: turn off database compression

jhult:


robinbb:
You should probably also remove the dependencies from the Cargo.toml by altering the features of the speedb package.

robinbb (approved):

robinbb:
This is mergeable, at your whim. But regardless of whether it is merged, we should know the effect upon performance and database space (even if only approximately).

robinbb:
Okay, I tested this on a database of 12,345 block height by timing 'just tier3'. Results:

Without it: took 4 min 51 s. DB/snapshot size was 4391 MB.
With this PR: took 6 min 31 s. DB/snapshot size was 20684 MB.

So, the database is 5 times bigger, and it's much slower. So, we should not do this because even if we add compression to the snapshot, everything is slower.

This was run on a machine with a very fast NVMe disk for the data, FYI.

jhult:
@robinbb, thanks for testing. I was trying to do so but ran out of disk space 🤦‍♂️.

PR #1250, opened 2024-07-06
Refactor cli - database commands

### LINKED ISSUE #1033: Problem: `server sync` is redundant, and should be removed to avoid confusion
'mina-indexer server sync' is the same as 'mina-indexer server start' if you have a database, but fails if you don't have a database. So, it is useless, and confusing. Remove 'sync'.

### LINKED ISSUE #1035: Problem: `server replay` is confusing; suggest flag on `server start`
Suggestion: get rid of the 'replay' subcommand and replace it with a flag on the 'start' command like "--self-check-on-start", which better reflects what happens.

### LINKED ISSUE #1203: Problem: scripts using mina-indexer must fork it only to wait for it
Scripts that use `mina-indexer` often want to have it ingest a bunch of block files in order to produce a working database. This includes our testing scripts. However, the indexer is nonresponsive during ingestion and the invoking script has no way to know that the Indexer has finished its ingestion other that to awkwardly poll for the existence of the Indexer's socket. This is a very poor user experience, and very error prone. (Is the right socket, or did another instance take over the socket? Did the indexer die, and it will never create the socket, causing infinite looping?) It requires writing more complex scripting logic than would be required to  fix this inside the Indexer by simply exiting after ingestion if a certain command line option is given.

This is causing complexity in the testing code, and slowing down the improvements to our CI system and our developer experience.

Not only must the polling be implemented, but the Indexer process must be forked to the background, and then waited for, with useless logging (the polling code) occupying the console when the user is actually interested in the Indexer's output/log.

Adding this feature would yield a lot of value, and it is easy to add.

Isaac-DeFrain:
- resolves #1033
- resolves #1035
- resolves #1203

### DIFF
--- a/ops/regression-test
+++ b/ops/regression-test
@@ -16,7 +16,6 @@ test_names = [
   'server_startup',
   'ipc_is_available_immediately',
   'startup_dirs_get_created',
-  'start_without_blocks_dir',
   'clean_shutdown',
   'clean_kill',
   'account_balance_cli',
--- a/ops/regression-test
+++ b/ops/regression-test
@@ -35,6 +35,8 @@ test_names = [
   'transactions',
   'snark_work',
   'snapshot',
+  'database_create',
+  'snapshot_database_dir',
   'rest_accounts_summary',
   'rest_blocks',
   'genesis_block_creator',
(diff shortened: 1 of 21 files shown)


robinbb (approved):
Very nice!!! This will break tier-3 CI, but I will merge this and then merge a fix for tier-3, later.

trevorbernard:
This also fixes the create database issue! Good work @Isaac-DeFrain

robinbb:
Reverted this, after all.

PR #1274, opened 2024-07-09
chore(Ruby): add standardrb and format Ruby code

jhult:
https://github.com/standardrb/standard

### DIFF
--- a/Justfile
+++ b/Justfile
@@ -46,6 +46,7 @@ lint:
   ruby -cw ops/download-staking-ledgers
   ruby -cw ops/stage-blocks
   ruby -cw ops/*.rb
+  standardrb --no-fix ops
   shellcheck tests/regression.bash
   @echo "--- Linting Rust code"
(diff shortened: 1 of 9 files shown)


jhult:
There is a [PR open](https://github.com/NixOS/nixpkgs/pull/310929) to add [`standardrb`](https://github.com/standardrb/standard) to [Nixpkgs](https://github.com/NixOS/nixpkgs), but it is not yet merged.

Thus, my current attempt is to add `standardrb` via a `gem`. However, that fails with this error:

```
ERROR: Could not find a valid gem 'standardrb' (>= 0), here is why: 
Unable to download data from https://rubygems.org/
Gem::Net::OpenTimeout: Failed to open TCP connection to rubygems.org:443 (execution expired)
```

robinbb:
I used 'Rubocop' to achieve same effect. Was available in recent nixpkgs.

PR #1280, opened 2024-07-09
chore(ops): add `.rb` extension to Ruby files

jhult:


robinbb (approved):

robinbb:
@jhult FYI, prior to this PR, the convention was that if there was no .rb extension, then the script was separately invocable (executable).

PR #1282, opened 2024-07-10
Optionally start block and staking ledger watchers

trevorbernard:
* Change default DB location to be more intune


trevorbernard:
I tried running tier1 locally but I'm unconvinced the code was causing my local build to break. Verifying on buildkite

Issue #1086, opened 2024-06-07
Problem: tier-2 testing does not include a straightforward load test


robinbb:

robinbb:
@jhult Would it be possible to easily construct a load test by invoking all of the HURL tests, but in parallel all at once?

PR #1296, opened 2024-07-15
feat: add (simple) load test to tier2 testing

jhult:
Fixes Granola-Team/mina-indexer#1086


jhult:
`test_hurl()` was already using `--parallel`. As part of this PR, I disabled that so that only `test_load()` uses `--parallel`.

Isaac-DeFrain (approved):
LGTM

Issue #767, opened 2024-04-09
Indexer does not always shutdown cleanly


robinbb:
SIGTERM is not sufficient to kill Indexer.

jhult:
https://github.com/Finomnis/tokio-graceful-shutdown seems like it would help.

jhult:
Related to #1144

robinbb:
The test is still commented out in 'regression.bash', and this is still not solved. Not sure why the issue was closed. When the Indexer is killed with SIGINT/SIGTERM, it fails to handle the socket.

PR #1299, opened 2024-07-16
fix(test): fix CLI shutdown

jhult:
Fixes Granola-Team/mina-indexer#767

The wrong PID was being captured. The PID captured was for a bash process and thus the kill signal wasn't reaching the Indexer itself.


robinbb:
@jhult It is true that the PID captured is not the PID of the indexer, but it is a sufficiently good PID to keep because sending it a signal will kill its child (the mina-indexer) as desired. The solution you have here will not work as-is because it backgrounds the 'idxr' command, but that command must not always be backgrounded as it is often used for foreground things.

robinbb (approved):

PR #1302, opened 2024-07-16
Add Granola development practices and principles doc

trevorbernard:
@Isaac-DeFrain @robinbb @jhult @n1tranquilla Take some time to think about and comment. Leaving in draft.


robinbb:
I wouldn't call this "coding guidelines", as it contains advice that is not what one normally finds in such a document. However, I'm fine to check this in somewhere with a different title.

trevorbernard:
> I wouldn't call this "coding guidelines", as it contains advice that is not what one normally finds in such a document. However, I'm fine to check this in somewhere with a different title.

Suggest an alternate title and it will be

PR #1326, opened 2024-07-23
Feat: GQL block producer count

### LINKED ISSUE #1321: Solution for indexing unique block producers for the last 1000 blocks
This stat must be:
* current with the latest tip (most recent block ingested)
* cheap to query

Isaac-DeFrain:
- resolves #1321


trevorbernard:
I think we should re-evaluate where we return these projection/views from. This relates to blocks but IMHO shouldn't be part of the blocks endpoint. But I will merge this PR

Issue #1340, opened 2024-07-30
Don't use saturating subtraction when checking for account creation levy fee for Fee Transfers


trevorbernard:
Subtle and hard to track bugs can occur. Better is to do explicit bounds checking to check for underflow scenarios.

PR #1341, opened 2024-07-30
Update account creation fee logic for CoinbaseFeeTransfers: avoid saturation subtraction

trevorbernard:
## Describe your changes

There are subtle bugs that can come from using saturating subtraction. When checking to see if the coinbase reward recipient is a new account, do explicit bounds checking to avoid underflow errors.

## Link issue(s) fixed

Fixes: https://github.com/Granola-Team/mina-indexer/issues/1340

### DIFF
--- a/rust/src/ledger/diff/account.rs
+++ b/rust/src/ledger/diff/account.rs
@@ -216,9 +216,9 @@ impl AccountDiff {
                                     ) => {
                                         if let Some(fee) = cb.t.fee_transfer_receiver_balance {
                                             let fee_transfer_receiver_balance = fee.t.t.t;
-                                            if (*total_fee)
-                                                .saturating_sub(MAINNET_ACCOUNT_CREATION_FEE.0)
-                                                == fee_transfer_receiver_balance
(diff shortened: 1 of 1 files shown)


rust/src/ledger/diff/account.rs
```diff
@@ -216,9 +216,9 @@ impl AccountDiff {
                                     ) => {
                                         if let Some(fee) = cb.t.fee_transfer_receiver_balance {
                                             let fee_transfer_receiver_balance = fee.t.t.t;
-                                            if (*total_fee)
-                                                .saturating_sub(MAINNET_ACCOUNT_CREATION_FEE.0)
-                                                == fee_transfer_receiver_balance
+                                            if *total_fee >= MAINNET_ACCOUNT_CREATION_FEE.0
```
robinbb, line 219: This implies that one cannot transfer less than the account creation fee. However, that is only true when the account has not yet been created. Is that true, here?

robinbb (changes requested):

robinbb (approved):

PR #1357, opened 2024-08-09
Represent `Coinbase.fee_transfer()` as an `Option<CoinbasePaymentDiff>` instead of `Vec<PaymmentDiff>`

n1tranquilla:


n1tranquilla:
I recommend hiding the whitespace when reviewing.

PR #1360, opened 2024-08-12
Code improvements to `InternalCommand::from_precomputed`

n1tranquilla:

### DIFF
--- a/rust/src/command/internal/mod.rs
+++ b/rust/src/command/internal/mod.rs
@@ -117,23 +117,47 @@ impl InternalCommand {
                             )
                         };
 
-                    internal_cmds.push(Self::FeeTransfer {
-                        sender: ic_sender.clone(),
-                        receiver: ic_receiver.clone(),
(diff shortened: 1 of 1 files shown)


rust/src/command/internal/mod.rs
```diff
@@ -76,74 +76,53 @@ impl InternalCommand {
...
+                                this_account_diff_fee.public_key.clone(),
+                                next_account_diff_fee.public_key(),
+                            )
+                        };
+
+                    internal_cmds.push(Self::FeeTransfer {
+                        sender: ic_sender.clone(),
+                        receiver: ic_receiver.clone(),
+                        amount: this_account_diff_fee.amount.0
+                            + next_account_diff_fee.amount().try_into().ok().unwrap_or(0),
```
trevorbernard, line 124: This is highly suspect. When dealing with token amounts we should never modify the intended value. Modifying the value is a bug. If we can't convert, throw a noisy Error

Issue #1352, opened 2024-08-08
`InternalCommand::from_precomputed` is hard to understand and might be simplified


n1tranquilla:
https://github.com/Granola-Team/mina-indexer/blob/9354e9fdd82d9c71ccfb4ac9efad6f780d51c64a/rust/src/command/internal/mod.rs#L74

n1tranquilla:
This was reverted

PR #1361, opened 2024-08-12
Represent `AccountDiffs::from_block_fees` as a `Vec<Vec<AccountDiff>>`

n1tranquilla:
## Describe your changes
All account diffs are represented as an vector of all account diffs at the top level, while each inner vector represents accounts diffs that sum to zero (debits and credits). This will greatly simplify the identification of related `AccountDiff`s downstream.

## Link issue(s) fixed
Relates to (but does not close) #1352 

### DIFF
--- a/rust/src/command/internal/mod.rs
+++ b/rust/src/command/internal/mod.rs
@@ -72,7 +72,10 @@ impl InternalCommand {
     ///
     /// See `LedgerDiff::from_precomputed`
     pub fn from_precomputed(block: &PrecomputedBlock) -> Vec<Self> {
-        let mut account_diff_fees = AccountDiff::from_block_fees(block);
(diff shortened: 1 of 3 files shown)


robinbb:
@n1tranquilla Possible to add an assertion that verifies that the inner vecs do indeed sum to zero?

robinbb:
@n1tranquilla Please do not assign to multiple persons. Pick exactly one victim.

robinbb:
@n1tranquilla This change merits a database version bump.

Isaac-DeFrain (approved):
LGTM

trevorbernard:
I'm a little late to respond but better would have been to model the Vec<Vec<AccountDiff>> as something like `Vec<(AccountDiff, Accountdiff)>` or some FeeTransfer Journal Entry type. Vector<AccountDiff> is too permissive. It's also missing the most important part which is the assertion to check the balance sums to 0.

PR #1363, opened 2024-08-13
Process transfers as debit/credit pairs in `InternalCommand::from_precomputed` function 

n1tranquilla:
## Describe your changes
* The main algorithm is now `O(n)` instead of `O(n * log(n))` 
* It is easier to read as transfers are now processed in debit/credit pairs
* For balanced pairs, runtime asserts that debits and credits sum to zero
* A duplicate was found in the feetransfers GQL for the first 120 blocks. This was fixed as a result of the rewrite.

## Link issue(s) fixed
Closes: #1352

### DIFF
--- a/rust/src/command/internal/mod.rs
+++ b/rust/src/command/internal/mod.rs
@@ -5,7 +5,7 @@ use crate::{
     ledger::{coinbase::Coinbase, diff::account::*, public_key::PublicKey},
 };
 use serde::{Deserialize, Serialize};
-use std::fmt;
+use std::{collections::HashMap, fmt};
 
 #[derive(Debug, Clone, PartialEq, Eq, Deserialize, Serialize)]
 pub enum InternalCommandKind {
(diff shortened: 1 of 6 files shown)


n1tranquilla:
A comparison between the production and local instances on the feetransfers endpoint reveals that there is one duplicate entry in the production version, which is not present in mine. 

Screen Shot 2024-08-14 at 8 55 09 AM

This means I should update the feetransfers.hurl test

Isaac-DeFrain (commented):
Nice work!

robinbb:
> A comparison between the production and local instances on the feetransfers endpoint reveals that there is one duplicate entry in the production version, which is not present in mine.

@n1tranquilla It appears that the version on the left is more correct, no? It shows the breakdown of fees. Is that yours?

n1tranquilla:
@robinbb, per block, the current code rolls up to only one internal command per recipient. Is this the correct way? IDK.

Isaac-DeFrain:
Yes, we should aggregate on recipient only (instead of sender & recipient). I think this will require a slight modification to how we represent internal commands.

PR #1367, opened 2024-08-14
Refactor ledger storage

### LINKED ISSUE #1354: Write persisted staged ledger accounts to db CF instead of serializing the ledger hash map
Currently, we serialize the entire ledger hash map when we want to persist it so when we need to access even a single account, we are forced to deserialize the entire ledger hash map. We are doomed to this fate.

Instead what we should do is maintain either a `pk -> account` CF for each persisted staged ledger or a single `{state hash}{pk} -> account` CF for all persisted staged ledgers.

Isaac-DeFrain:
## Describe your changes

Refactor of the ledger/account stores to persist all ledgers as column families of accounts, instead of serializing/deserializing the entire ledger.

## Link issue(s) fixed

Resolves #1354


jhult:
@Isaac-DeFrain, [test failures here](https://buildkite.com/granola/mina-indexer-tier-1/builds/1928#01915c80-c167-44c7-80ab-eb79a53e3ef9/470-1202).

Isaac-DeFrain:
Opening a different PR instead...

PR #1368, opened 2024-08-14
Process transfers as debit/credit pairs in InternalCommand::from_precomputed function

n1tranquilla:
## Describe your changes
* `InternalCommand::from_precomputed` algorithm is greatly simplified and much easier to read

## Link issue(s) fixed
closes: #1352 

### DIFF
--- a/rust/src/block/precomputed.rs
+++ b/rust/src/block/precomputed.rs
@@ -203,8 +203,9 @@ impl PrecomputedBlock {
     }
 
     pub fn commands(&self) -> Vec<UserCommandWithStatus> {
-        let mut commands = self.commands_post_diff();
-        commands.append(&mut self.commands_pre_diff());
+        // This order is the correct order
+        let mut commands = self.commands_pre_diff();
+        commands.append(&mut self.commands_post_diff());
         commands
     }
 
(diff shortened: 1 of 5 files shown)


n1tranquilla:
After having a conversation with @Isaac-DeFrain on the origin of these tests, I have decided to repurpose the tests in this PR to serve as structural tests, and not assertions on the values within the response. He has indicated that he is unsure if these tests represent careful calculation over many blocks to arrive at the total number of internal commands.

Furthermore, the position of internal commands within the response structure has changed, since we are no longer aggregating internal commands _per recipient, per block_. This is a deviation from minaexplorer.com, but I think it more accurately represents fee transfers within a block.

trevorbernard:
>Furthermore, the position of internal commands within the response structure has changed, since we are no longer aggregating internal commands per recipient, per block. This is a deviation from minaexplorer.com, but I think it more accurately represents fee transfers within a block.

I don't agree with this conclusion because feetransfers are aggregated per public_key in the archive node and every other block explorer. Why would we deviate from this? The network doesn't issue multiple feetransfers to the same prover at a given block height.

n1tranquilla:
> The network doesn't issue multiple feetransfers to the same prover at a given block height.

If it is indeed the truth that the network doesn't issue multiple feetransfers to the same prover at a given block height, I will adjust this PR.

trevorbernard (approved):

PR #1372, opened 2024-08-15
Refactorings that move codebase closer to double entry accounting principles

### LINKED ISSUE #1366: `AccountDiff::from_command` does not represent accounts diffs as debit/credit pairs

n1tranquilla:
## Describe your changes
* Starting with `AccountDiff::from_command`, the rest of the code is refactored to support returning a list of double entries of debits and credits

## Link issue(s) fixed
* Closes #1366 

### DIFF
--- a/rust/src/ledger/diff/account.rs
+++ b/rust/src/ledger/diff/account.rs
@@ -83,40 +83,49 @@ pub enum AccountDiffType {
 }
 
 impl AccountDiff {
-    pub fn account_creation_payment_diff(new_account: &PublicKey) -> Self {
+    pub fn account_creation_payment_diff(new_account: &PublicKey) -> Vec<Self> {
         let public_key = new_account.clone();
-        AccountDiff::CreateAccount(PaymentDiff {
-            public_key,
-            update_type: UpdateType::Debit(None),
-            amount: Amount(MAINNET_ACCOUNT_CREATION_FEE.0),
-        })
+        vec![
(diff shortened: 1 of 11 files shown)


Isaac-DeFrain (commented):
LGTM

Isaac-DeFrain:
@n1tranquilla is this ready?

n1tranquilla:
@Isaac-DeFrain it is ready, but let's make tier2 passes on main successfully first before merging.

rust/src/ledger/diff/account.rs
```diff
@@ -83,40 +83,51 @@ pub enum AccountDiffType {
...
+                ),
+                update_type: UpdateType::Credit,
+                amount: Amount(MAINNET_ACCOUNT_CREATION_FEE.0),
+            }),
+            AccountDiff::CreateAccount(PaymentDiff {
+                public_key,
+                update_type: UpdateType::Debit(None),
+                amount: Amount(MAINNET_ACCOUNT_CREATION_FEE.0),
+            }),
+        ]
```
Isaac-DeFrain, line 101: This is definitely how I think about this creation fee (it is "burned"), but do we know that this is actually how it is? If it isn't, I'd argue that this is more confusing.

Isaac-DeFrain: This also leads to the question of when the burn address is added to the ledger...

PR #1383, opened 2024-08-17
Fix `amount()` function for the `AccountDiff::CreateAccount(PaymentDiff)` type

n1tranquilla:
## Describe your changes
* The `amount()` function properly handles returning the amount for the creation fee.


Isaac-DeFrain:
We will remove the `CreateAccount` account diffs and we don't currently use them directly to account for creation fees

PR #1385, opened 2024-08-18
Ensure that debit and credit amounts are properly issued when creating accounts

n1tranquilla:
## Describe your changes
* Debits and Credits for account creation were `u64` (positive). Now they are `i64` as debits are negative, and credits are positive.


n1tranquilla:
> I have verified new and existing tests pass locally with my changes.

I have run tier2 tests and they pass until `test_watch_staking_ledgers` which always fails for mac users.

PR #1386, opened 2024-08-19
Add tests for DBAccountUpdate

n1tranquilla:
## Describe your changes
* DBAccountUpdate will need to be refactored to handle `Vec<Vec<AccountDiff>>`; before doing so, we add tests.
* Some conditional logic to update HashMap from apply/unapply `Vec<AccountDiff>` is removed so that diffs are always applied/unapplied. 

## Link issue(s) fixed
* Closes: https://github.com/Granola-Team/mina-indexer/issues/1387

### DIFF
--- a/rust/src/ledger/store/best.rs
+++ b/rust/src/ledger/store/best.rs
@@ -105,58 +105,54 @@ impl DBAccountUpdate {
                     );
                 }
                 Coinbase(diff) => {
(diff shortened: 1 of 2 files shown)


n1tranquilla:
> I have verified new and existing tests pass locally with my changes.

I have run tier2 tests and they pass until test_watch_staking_ledgers which always fails for mac users.

n1tranquilla:
This code is actually not used 😦

PR #1410, opened 2024-08-22
Add useful script

n1tranquilla:
## Describe your changes
I found this script useful when I wanted to see if a block is canonical or non-canonical. It will traverse files in a directly printing out each parent until no parent is found. 


robinbb (commented):
The 'ops' directory is the right place for this.

Isaac-DeFrain (dismissed):
Please move to ops/

robinbb (approved):

PR #1411, opened 2024-08-22
feat: Optionally ingest orphan blocks

Isaac-DeFrain:
## Describe your changes

- introduces the CLI flag `--do-not-ingest-orphan-blocks` and associated functionality
- changes `deploy.rb` to use the flag for tier3 testing
- adds delegate field to tier3 ledger diff comparison

## Link issue(s) fixed

- #1404


ops/deploy.rb
```diff
@@ -36,7 +36,8 @@
     '--ledger-cadence', '5000',
     '--database-dir', db_dir(BLOCKS_COUNT),
     '--blocks-dir', blocks_dir(BLOCKS_COUNT),
-    '--staking-ledgers-dir', LEDGERS_DIR
+    '--staking-ledgers-dir', LEDGERS_DIR,
+    '--do-not-ingest-orphan-blocks'
```
n1tranquilla, line 40: `--canonical-only` would also be good

PR #1415, opened 2024-08-26
Rewrite discovery algorithm

n1tranquilla:


jhult:
We bumped the database version in this PR (which is the correct thing to do).

@Isaac-DeFrain's recent commit to `main` (https://github.com/Granola-Team/mina-indexer/commit/ed672d86f0ea261e934b205c0ddc067502347daa) also bumped to the same database version. This may cause issues with CI.

jhult:
LGTM but our goal is to pass `tier 3` CI.

Merging but I know that we may need to revert.

PR #1420, opened 2024-08-27
Introduce `SignedAmount` without using it yet

n1tranquilla:
## Describe your changes
With a `SignedAmount` as a new field in the `PaymentDiff`, we can handle negative balances in internal commands


jhult (approved):
Nice work!

PR #1422, opened 2024-08-27
Remove `saturating_sub` on Amount as balances cannot be negative

n1tranquilla:


Isaac-DeFrain (approved):
YOLO

PR #1425, opened 2024-08-29
Separate block iteration from block parsing

n1tranquilla:
## Describe your changes
* Block iteration and block parsing are currently entwined concepts.
* Currently, each iteration of the block paths results in parsing the file contents into a `PrecomputedBlock`. 
* This means that entire JSON of each block is serialized into memory upon each iteration, making ingestion slow. 
* In the future, we may want to avoid serializing the entire file into memory when ingesting. This helps.

### DIFF
--- a/rust/src/state/mod.rs
+++ b/rust/src/state/mod.rs
@@ -501,7 +501,7 @@ impl IndexerState {
                         "Finished ingesting and applying {} blocks ({}) to the witness tree in {:?}",
                         self.blocks_processed,
                         bytesize::ByteSize::b(self.bytes_processed),
-                        total_time.elapsed() + offset,
+                        pretty_print_duration(total_time.elapsed() + offset),
                     );
                     break;
                 }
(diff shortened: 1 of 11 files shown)


rust/src/state/mod.rs
```diff
@@ -476,48 +476,42 @@ impl IndexerState {
...
+                            self.add_block_to_store(&block, block_bytes, true)?;
                         }
                     }
                 }
+                Ok(None) => {
+                    info!(
+                        "Finished ingesting and applying {} blocks ({}) to the witness tree in {:?}",
+                        self.blocks_processed,
+                        bytesize::ByteSize::b(self.bytes_processed),
+                        total_time.elapsed() + offset,
```
jhult, line 504:
```suggestion
                        pretty_print_duration(total_time.elapsed() + offset),
```

jhult:
@n1tranquilla,

1. Does this need a version bump?
2. Do you have any metrics to determine the impact on performance?

n1tranquilla:
It should require a version bump, and there ought to be no impacts on performance. It's just a refactor that will facilitate future refactorings.

jhult (approved):
Let's go ahead and merge this.

PR #1430, opened 2024-08-29
more robust previous_state_hash parsing

n1tranquilla:

### DIFF
--- a/rust/src/block/previous_state_hash.rs
+++ b/rust/src/block/previous_state_hash.rs
@@ -1,6 +1,5 @@
 use crate::block::BlockHash;
 use serde::{Deserialize, Serialize};
-use serde_json::Value;
 use std::{
     fs::File,
     io::{BufReader, Read},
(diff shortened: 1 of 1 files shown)


n1tranquilla:
The database has already been incremented in previous PR. If tier3 does not rebuild the database, we'll have to incr the version.

rust/src/block/previous_state_hash.rs
```diff
@@ -23,31 +22,25 @@ pub struct PreviousStateHash(pub String);
...
+
+        // Locate "previous_state_hash" within the buffer
+        let prev_state_hash_key = "\"previous_state_hash\"";
+        if let Some(hash_pos) = buffer.find(prev_state_hash_key) {
+            let hash_start = hash_pos + prev_state_hash_key.len();
+            // Find the first quote after the colon
+            if let Some(quote_start) = buffer[hash_start..].find('"') {
+                let actual_start = hash_start + quote_start + 1; // Move to the start of the hash
+                                                                 // Find the ending quote
+                if let Some(quote_end) = buffer[actual_start..].find('"') {
```
trevorbernard, line 41: The hash is fixed size so you shouldn't need to find the closing double quote

trevorbernard: I believe it's 52 characters wide

jhult: Will try to address this in a follow-up PR.

jhult: Fixed in this commit: https://github.com/Granola-Team/mina-indexer/commit/931f86f52bb55891aba064afbf302bdefc8d8d2f

PR #1439, opened 2024-09-04
Use memory on stack when extracting height and state hash from filename

n1tranquilla:
## Describe your changes
Allocates memory on the stack (not the heap) when parsing height and state hash from fn.


rust/src/block/mod.rs
```diff
@@ -431,24 +417,18 @@ mod tests {
 
     #[test]
     fn extract_state_hash_test() {
-        let filename1 =
```
n1tranquilla, line 434: I'm 99% certain this is not a valid filename anymore.

n1tranquilla: @Isaac-DeFrain ?

jhult (approved):
LGTM

Issue #1442, opened 2024-09-04
Various functions allocated memory on the heap when they could allocated memory on the stack


n1tranquilla:
All of these return a vector of bytes (`Vec<u8>`) when they could return an array of known size.

```
staged_account_balance_sort_key
block_height_key
block_global_slot_key
pk_block_sort_key
txn_sort_key
pk_txn_sort_key
block_txn_index_key
txn_block_key
staged_ledger_account_balance_iterator
staking_ledger_epoch_key_prefix
pk_txn_sort_key_prefix
staking_ledger_account_key
```

Other functions affected to bytes are being converted .to_vec()
```
add_pk_delegate
get_pk_username
update_usernames
```

n1tranquilla:
Significant progress has been made on this already.

PR #1443, opened 2024-09-04
`BlockHash::to_bytes -> [u8; BlockHash::LEN]` instead of `... -> Vec<u8>`

n1tranquilla:
## Describe your changes
* Makes a step towards better memory utilization. 
* Follow up issue has been created to convert other functions to &[u8] : https://github.com/Granola-Team/mina-indexer/issues/1442

### DIFF
--- a/rust/src/store/block_store_impl.rs
+++ b/rust/src/store/block_store_impl.rs
@@ -991,14 +991,14 @@ impl BlockStore for IndexerStore {
 /// `{block height BE}{state hash}`
 fn block_height_key(block: &PrecomputedBlock) -> Vec<u8> {
     let mut key = to_be_bytes(block.blockchain_length());
-    key.append(&mut block.state_hash().to_bytes());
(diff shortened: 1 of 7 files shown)


rust/src/store/block_store_impl.rs
```diff
@@ -991,22 +991,22 @@ impl BlockStore for IndexerStore {
 /// `{block height BE}{state hash}`
 fn block_height_key(block: &PrecomputedBlock) -> Vec<u8> {
     let mut key = to_be_bytes(block.blockchain_length());
-    key.append(&mut block.state_hash().to_bytes());
+    key.append(&mut block.state_hash().to_bytes().to_vec());
```
jhult, line 994: If we almost always have to convert it to a `Vec`, there might not be much savings by avoiding a `Vec` to begin with.

jhult (approved):
LGTM

PR #1489, opened 2024-09-11
(devops) use exit code of diff command to fail tier3

trevorbernard:
## Describe your changes

- warn user of regression when diff command has output and return the proper exit code


n1tranquilla:
Thanks @trevorbernard I was wondering why it was failing.

Issue #1353, opened 2024-08-08
`TransactionQueryInput` query crashes for some `dateTimeGte` and `limit` combinations


n1tranquilla:
# Case A: Crashing case

Query:
```
query TransactionsQuery($limit: Int, $sort_by: TransactionSortByInput!, $query: TransactionQueryInput!) {
  transactions(limit: $limit, sortBy: $sort_by, query: $query) {
    hash
    amount
    kind
    blockHeight
    block {
      dateTime
    }
  }
}
```

Variables:
```
{
  "limit": 10000,
  "sort_by": "BLOCKHEIGHT_DESC",
  "query": {
    "canonical": true,
    "kind": "PAYMENT",
    "dateTimeGte": "2024-06-03T00:00:00.000Z"
  }
}
```

Result:
```
{
  "errors": [
    {
      "message": "Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON",
      "stack": "SyntaxError: Unexpected token '<', \"<!DOCTYPE \"... is not valid JSON"
    }
  ]
}
```

# Case B: Non-Crashing case

Query: (same)

Variables:
```
{
  "limit": 10000,
  "sort_by": "BLOCKHEIGHT_DESC",
  "query": {
    "canonical": true,
    "kind": "PAYMENT",
    "dateTimeGte": "2024-06-02T00:00:00.000Z"
  }
}
```

n1tranquilla:
Blocks: https://github.com/Granola-Team/mina-block-explorer/issues/986

Isaac-DeFrain:
We'll need to deploy a recent build to prod so I can check this

Isaac-DeFrain:
Related #1539

PR #1494, opened 2024-09-11
fix: Use millis over internal string for ordering GQL date time

Isaac-DeFrain:
## Link issue(s) fixed

- #1353


jhult (approved):
Nice!

PR #1497, opened 2024-09-12
Make canonical discovery algorithm near instantaneous

n1tranquilla:
## Describe your changes
* Build a map of previous state hashes using threads to avoid blocking I/O on single thread.
* Benchmark on `40k` blocks yield `0s` discovery vs `19s` on main.
* Given that this is a 10 minute process in tier3, this should yield savings.


Isaac-DeFrain (approved):
Sweet!

Issue #1339, opened 2024-07-29
Problem: this query takes way too long


n1tranquilla:
(Edited by @robinbb)

This query pins 2 CPUs at 100% and does not return before Cloudflare times out the request (1 minute?).

So, it is unacceptably slow because the front-end seems to hang.

```
[...]

PR #1508, opened 2024-09-13
Possible fix for performance issue with accounts GQL query

n1tranquilla:
## Describe your changes
* Instead of loading all internal commands into memory and then truncating, load only the limit
* There are several iterations taken per internal command loaded into memory, so this should have a cascading effect (in a good way)

### DIFF
--- a/rust/src/web/graphql/feetransfers/mod.rs
+++ b/rust/src/web/graphql/feetransfers/mod.rs
@@ -253,8 +253,8 @@ impl FeetransferQueryRoot {
         // TODO
         // recipient query
         if let Some(recipient) = query.as_ref().and_then(|q| q.recipient.clone()) {
-            let mut fee_transfers: Vec<FeetransferWithMeta> = db
-                .get_internal_commands_public_key(&recipient.into())?
+            let fee_transfers: Vec<FeetransferWithMeta> = db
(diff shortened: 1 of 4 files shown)


rust/src/web/graphql/feetransfers/mod.rs
```diff
@@ -253,8 +253,8 @@ impl FeetransferQueryRoot {
         // TODO
         // recipient query
         if let Some(recipient) = query.as_ref().and_then(|q| q.recipient.clone()) {
-            let mut fee_transfers: Vec<FeetransferWithMeta> = db
-                .get_internal_commands_public_key(&recipient.into())?
+            let fee_transfers: Vec<FeetransferWithMeta> = db
+                .get_internal_commands_public_key(&recipient.into(), limit as u32)?
```
jhult, line 257: Can you limit this early? Don't you need to wait for all other filters to be complete?

n1tranquilla: I'm going to resolve this for now. If there is more improvement that can be made, let's follow up with another PR.

jhult (approved):
LGTM

jhult:
This may resolve #1339

PR #1509, opened 2024-09-13
Performance fix account GraphQL endpoint

n1tranquilla:
## Describe your changes
* Strategy is to iterate and filter over internal commands until there are 'enough'
* Iteration favours lower block heights first. This could be tweaked to favour higher blocks first, if desired


jhult (approved):
Nice code comments and thanks for the `hurl` test.

jhult:
Possible fix for #1339

PR #1511, opened 2024-09-14
refactor: Store keys

Isaac-DeFrain:
## Describe your changes

This PR refactors the store key helper functions.

## Link issue(s) fixed

- #1442


n1tranquilla:
I think this is risky having so many changes in one PR, however I also want to see this change go through. It is very likely that any issue would be caught in the unit tests first, as the move towards specificity (array of bytes of known size) easily panics if incorrect.

PR #1521, opened 2024-09-16
chore: upgrade nix

jhult:


n1tranquilla:
It would be helpful to has `standardrb` linting in another commit

PR #1540, opened 2024-09-18
Add two ruby scripts that I use

n1tranquilla:
## Describe your changes
* One removes "proofs" from PCBs 
* One recycles existing PCB files when preparing a new deployment  


Isaac-DeFrain (approved):
LGTM

ops/remove-proofs-from-pcbs.rb
```diff
@@ -0,0 +1,47 @@
...
+
+# Recursive method to remove all "proofs" properties from a JSON structure
+def remove_proofs(obj)
+  if obj.is_a?(Hash)
+    obj.delete("proofs")
+    obj.each { |key, value| remove_proofs(value) }
+  elsif obj.is_a?(Array)
+    obj.each { |item| remove_proofs(item) }
+  end
+end
```
jhult, line 14: Do you know if this is slower than shelling out to `jq`? We already use `jq` for other tests.

jhult (approved):

ops/remove-proofs-from-pcbs.rb
n1tranquilla: I can't tell. It seemed very fast for me. Just a few seconds for 5000 blocks

PR #1560, opened 2024-09-23
fix(store): snark updates

Isaac-DeFrain:
## Describe your changes

Fixes the SNARK store updates to occur on best block updates.

## Link issue(s) fixed

- #1555


Isaac-DeFrain:
Unapplying SNARK updates will be cover in a follow-up PR

PR #1566, opened 2024-09-24
fix(gql): correct epoch SNARK fees

Isaac-DeFrain:
## Describe your changes

Corrects the epoch SNARK fee values stored in the db and used by GQL.

## Link issue(s) fixed

- #1565


n1tranquilla (commented):
Happy to merge, however we don't have any tests to prove this will work. We do have known wrong cases [here](http://127.0.0.1:5172/analytics/snarker-leaderboard?epoch=78), however, that you could put in the hurl.

jhult (approved):
LGTM

PR #1577, opened 2024-09-26
fix(tests): regression suite should always use same build type

jhult:
## Describe your changes

Fix inconsistent build type


Isaac-DeFrain (approved):
Is it a feature or a bug to not use the nix binary for CI tier 2? I recently added the ability to choose between debug and nix for local development. I’ll approve, just wanted to point that out

Issue #1623, opened 2024-10-12
Analytics query fails and crashes indexer


n1tranquilla:
Query:
```
query InternalCommands($limit: Int, $sort_by: FeetransferSortByInput!, $query: FeetransferQueryInput!) {
  feetransfers(limit: $limit, sortBy: $sort_by, query: $query) {
    fee
    blockHeight
  }
}
```

Variables:
```
{
  "limit": 10000000,
  "sort_by": "BLOCKHEIGHT_DESC",
  "query": {
    "canonical": true,
    "blockHeight_gte": 358604,
    "blockHeight_lte": 359604
  }
}
```

Isaac-DeFrain:
I’ll look into this

n1tranquilla:
The limit is too high. I didn't think it mattered. But I think we should bound the limit based on the two heights, if they are both supplied.

n1tranquilla:
I'll work around this issue in the front end by lowering the limit.

n1tranquilla:
This is still an issue.

robinbb:
No longer an issue.

PR #1624, opened 2024-10-12
If block height range supplied, and block is out of bounds, exit early

n1tranquilla:
## Describe your changes
Relies on query to determine an effective `limit`

## Link issue(s) fixed
Closes: #1623


Isaac-DeFrain (approved):
LGTM

PR #1721, opened 2025-01-27
feat(genesis): use full hardfork genesis ledger

Isaac-DeFrain:
## Describe your changes

- uses full hardfork genesis ledger
- adds v1/v2 genesis ledger & state functions
- improves git lfs file management


rust/src/utility/compression.rs
```diff
@@ -0,0 +1,17 @@
...
+}
+
+/// Decompress gzip compressed bytes
+pub(crate) fn decompress_gzip(bytes: &[u8]) -> anyhow::Result<Vec<u8>> {
+    let mut buf = vec![];
+    let mut gz = GzDecoder::new(bytes);
+
+    gz.read_to_end(&mut buf)?;
+    Ok(buf)
+}
```
jhult, line 17: (not-blocking; FWIW): there is already code to `tar` a file.

https://github.com/Granola-Team/mina-indexer/blob/7f3efe4cb7b22f811cf5d2b005acb5f2f6331e31/rust/src/store/mod.rs#L333-L363

Isaac-DeFrain: Yes, but we need to compress the file, not tar it

jhult (approved):

Issue #1813, opened 2025-03-20
Convert from `just` to `rake`


jhult:

jhult:
Related to #1812

jhult:
It should test using a binary and (thus) not use any internal functions.

PR #1823, opened 2025-03-26
feat(ops): convert `just` to `rake`

jhult:
## Describe your changes

Converting from `just` to `rake`

## Link issue(s) fixed

Resolves Granola-Team/mina-indexer#1813


jhult:
Tier-3 also passed on my local. I'm going to merge this and see what happens in CI. If there are issues, we can always leave `just` in place for Tier-3 (although it doesn't seem to be used).

PR #1930, opened 2026-05-30
feat(cli): add blocks compare subcommand

### LINKED ISSUE #1606: Add CLI `blocks compare state_hash1 state_hash2` utility

RyanJamesStewart:
## Summary

Implements `blocks compare` as requested in #1606.

```
mina-indexer blocks compare --state-hash <A> --other-state-hash <B> [--path <FILE>]
```

It reports the better of two blocks according to consensus ordering (longer chain, then VRF / state-hash tie-break). The handler reuses the existing `BlockStore::block_cmp`, so the consensus rule is not reimplemented here. `--path` writes the winning state hash to a file instead of stdout.

Behaviour:
- both hashes present: prints the better state hash (symmetric under argument order)
- one or both missing from the store: prints `Block missing from store: ...`
- malformed hash: prints `Invalid state hash: ...`

## Verification

Built and tested against a local indexer store built from the repo's own block fixtures (`canonical_chain_discovery`), with two competing height-10 blocks ingested:

- same-height fork: returns the better block, and the result is identical when the two arguments are swapped
- different heights: the longer-chain block wins, in both argument orders
- missing-from-store and invalid-hash paths produce the messages above
- `--path` writes the winning hash to the given file

A `blocks compare --help` assertion was added to `tests/regression.bash` alongside the other subcommand help checks. `cargo clippy` is clean.

Closes #1606

### DIFF
--- a/rust/src/client.rs
+++ b/rust/src/client.rs
@@ -174,6 +174,21 @@ pub enum Blocks {
         #[arg(long, default_value_t = false)]
         verbose: bool,
     },
+
+    /// Compare two blocks and report the better state hash
+    Compare {
+        /// First state hash to compare
+        #[arg(long)]
+        state_hash: String,
+
+        /// Second state hash to compare
+        #[arg(long)]
+        other_state_hash: String,
+
+        /// Path to write the result [default: stdout]
+        #[arg(long)]
+        path: Option<PathBuf>,
+    },
 }
 
 #[derive(Subcommand, Debug, Encode, Decode)]
--- a/rust/src/unix_socket_server.rs
+++ b/rust/src/unix_socket_server.rs
@@ -470,6 +470,46 @@ pub async fn handle_connection(
                     }
                 }
             }
+            ClientCli::Blocks(Blocks::Compare {
+                state_hash,
+                other_state_hash,
+                path,
+            }) => {
+                debug!("Received block-compare command for {state_hash} and {other_state_hash}");
+                if !StateHash::is_valid(&state_hash) {
+                    invalid_state_hash(&state_hash)
+                } else if !StateHash::is_valid(&other_state_hash) {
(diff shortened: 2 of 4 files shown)


robinbb:
@RyanJamesStewart You, again! You seem to be following my work, despite the work not being related (except by me). How did you decide to work on this? Was it related to my willingness to interact with you on the Dune work?

Regardless, I can't approve this as-is due to lack of regression tests.

RyanJamesStewart:
Added in 9f573cd. `test_blocks_compare` follows the existing `test_block_children` pattern (same `stage_blocks v1 10` fixture set) and covers every behaviour documented in the PR:

- different heights → longer chain wins (symmetric under arg swap)
- same-height fork → canonical sibling wins (symmetric under arg swap)
- self-compare returns the input hash
- well-formed but absent hash → `Block missing from store: ...`
- malformed hash → `Invalid state hash: ...`
- `--path` writes the winning state hash to the given file

Wired into the dispatcher in `tests/regression.bash` and the `test_names` list in `tests/regression-test.rb`. Clean under `bash -n` and `shellcheck`. The four non-fixture branches were also smoke-tested against a live indexer to confirm the assertion strings match the server's output verbatim.

robinbb:
@RyanJamesStewart "Usage: mina-indexer blocks compare" is not the correct usage message.

RyanJamesStewart:
Fixed in `afce86ad`. The help assertion now matches the complete usage line clap emits, not just the leading prefix:

```
Usage: mina-indexer blocks compare [OPTIONS] --state-hash <STATE_HASH> --other-state-hash <OTHER_STATE_HASH>
```

Matched with `grep -iqF` against the live `blocks compare --help` output (fixed-string, since `[OPTIONS]` would otherwise be read as a bracket expression). Clean under `bash -n` and `shellcheck`.

PR #1931, opened 2026-06-04
perf(store): cache block sort keys instead of re-reading per compare

RyanJamesStewart:
## Summary

The four `sort_by` sites in `BlockStore` ([`get_blocks_at_height`](https://github.com/Granola-Team/mina-indexer/blob/main/rust/src/store/block_store_impl.rs#L628), [`get_blocks_at_slot`](https://github.com/Granola-Team/mina-indexer/blob/main/rust/src/store/block_store_impl.rs#L679), [`get_blocks_at_public_key`](https://github.com/Granola-Team/mina-indexer/blob/main/rust/src/store/block_store_impl.rs#L730), [`get_block_children`](https://github.com/Granola-Team/mina-indexer/blob/main/rust/src/store/block_store_impl.rs#L744)) use a comparator that reads each block's canonicity and `BlockComparison` from RocksDB on every pairwise comparison: 6 column-family reads and 2 `serde_json::from_slice` calls per cmp, evaluated O(N log N) times.

Replace the free `block_cmp` function with `block_sort_key`, returning a `(Reverse<bool>, BlockComparison)` key, and switch the four call sites to `sort_by_cached_key`. DB reads and deserializations drop from O(N log N) to O(N) per sort; key comparisons are then in-memory.

## Result

Microbenchmark on a real mainnet 7-block fork (`tests/data/sequential_blocks`, height 105496), 10 000 iterations per run, three runs each:

| Build | Mean per call |
|---|---|
| Before (`sort_by` + `block_cmp`) | **29.15 µs** |
| After (`sort_by_cached_key` + `block_sort_key`) | **6.30 µs** |

**4.6x faster.** Variance under 5% across runs. The win scales: at fork width N, comparator-side DB work goes from `6 N log N` reads / `2 N log N` deserializes to `6 N` / `2 N`.

## Behavioural note

The previous comparator's `(Some(Canonical), Some(Canonical))` arm returned `Ordering::Less` regardless of argument order. That is an invariant-violating state (at most one canonical block per height) but technically not a total order. The new key gives a well-defined deterministic ordering in that case.

## Test plan

- [x] `cargo check --release` clean
- [x] `cargo clippy --release --all-targets --all-features -- -D warnings` clean
- [x] 226 unit tests pass (`cargo test --release --lib`)
- [x] 107 integration tests pass (`cargo test --release --test mod`), including all 8 canonicity tests and all 7 block-store tests, which exercise the affected sort

### DIFF
--- a/rust/src/store/block_store_impl.rs
+++ b/rust/src/store/block_store_impl.rs
@@ -625,7 +625,7 @@ impl BlockStore for IndexerStore {
                 Some(bytes) => blocks.push(StateHash::from_bytes(&bytes)?),
             }
         }
-        blocks.sort_by(|a, b| block_cmp(self, a, b));
+        blocks.sort_by_cached_key(|sh| block_sort_key(self, sh));
         Ok(blocks)
     }
 
@@ -676,7 +676,7 @@ impl BlockStore for IndexerStore {
             }
         }
 
-        blocks.sort_by(|a, b| block_cmp(self, a, b));
+        blocks.sort_by_cached_key(|sh| block_sort_key(self, sh));
         Ok(blocks)
     }
 
@@ -727,7 +727,7 @@ impl BlockStore for IndexerStore {
             }
         }
 
-        blocks.sort_by(|a, b| block_cmp(self, a, b));
+        blocks.sort_by_cached_key(|sh| block_sort_key(self, sh));
         Ok(blocks)
     }
 
@@ -741,7 +741,7 @@ impl BlockStore for IndexerStore {
                     self.get_block_parent_hash(b).ok().flatten() == Some(state_hash.clone())
                 })
                 .collect();
-            children.sort_by(|a, b| block_cmp(self, a, b));
+            children.sort_by_cached_key(|sh| block_sort_key(self, sh));
             return Ok(children);
         }
         bail!("Block missing from store {state_hash}")
@@ -1925,19 +1925,17 @@ fn is_genesis_hash(hash: &StateHash) -> bool {
     hash.0 == MAINNET_GENESIS_HASH || hash.0 == HARDFORK_GENESIS_HASH
 }
 
-fn block_cmp(db: &IndexerStore, a: &StateHash, b: &StateHash) -> std::cmp::Ordering {
-    use std::cmp::Ordering;
+// Canonical blocks sort before non-canonical ones; otherwise BlockComparison
+// orders by (length, vrf, state hash) per consensus.
+type BlockSortKey = (std::cmp::Reverse<bool>, BlockComparison);
 
-    let a_canonicity = db.get_block_canonicity(a).ok().flatten();
-    let b_canonicity = db.get_block_canonicity(b).ok().flatten();
-    let a_cmp = db.get_block_comparison(a).unwrap().unwrap();
-    let b_cmp = db.get_block_comparison(b).unwrap().unwrap();
-
-    match (a_canonicity, b_canonicity) {
-        (Some(Canonicity::Canonical), _) => Ordering::Less,
-        (_, Some(Canonicity::Canonical)) => Ordering::Greater,
-        _ => a_cmp.cmp(&b_cmp),
-    }
+fn block_sort_key(db: &IndexerStore, state_hash: &StateHash) -> BlockSortKey {
+    let is_canonical = matches!(
+        db.get_block_canonicity(state_hash).ok().flatten(),
+        Some(Canonicity::Canonical)
+    );
+    let comparison = db.get_block_comparison(state_hash).unwrap().unwrap();
+    (std::cmp::Reverse(is_canonical), comparison)
 }
 
 fn display_mode(mode: IteratorMode) -> String {


robinbb:
@RyanJamesStewart This is great! What a horrible inefficiency you are correcting. Did you run the Bash-file-driven test suite? (Your plan suggests not.) How many callers of the sort_by in BlockStore are there?

RyanJamesStewart:
Thanks. Two answers.

There are 4 callers. All four `sort_by` sites in `BlockStore` (`get_blocks_at_height`, `get_blocks_at_slot`, `get_blocks_at_public_key`, `get_block_children`) are converted to `sort_by_cached_key`, and the free `block_cmp` function is removed with no remaining callers. The `BlockStore::block_cmp` trait method is a separate function (the pairwise compare used by the `blocks compare` subcommand), so it is intentionally left untouched.

On the speedup, it scales with fork width the way the O(N log N) to O(N) read count predicts. Measuring the comparator cost in isolation across every multi-block fork in `tests/data/sequential_blocks` (release build, best of three runs of 10k iterations each), the before/after per-call cost goes from 3.48 to 2.58 µs at width 3 (1.4x), 7.48 to 4.29 µs and 10.54 to 3.81 µs at width 4 (1.7x and 2.8x), 20.89 to 5.49 µs at width 6 (3.8x), and 37.35 to 6.45 µs at the 7-block fork (5.8x). The before side grows super-linearly while the after side stays roughly flat. These isolate the sort itself with no height-index lookup, which is why the before figures run a little higher than the end-to-end `get_blocks_at_height` number in the PR body; the after figure reproduces it.

On testing, I ran the bash-driven regression tests that drive these sort sites (`blocks`, `best_chain_v1`, `canonical_root`, `canonical_threshold`) and they pass, and the fork-width ordering is covered by the lib and integration tests in the PR's test plan. The fork-staging tests pull fixtures from the blocks bucket, which I am not set up to fetch locally, so I left those for your side.

Worth flagging on the behavioural note: the old comparator's `(Some(Canonical), Some(Canonical))` arm returned `Ordering::Less` regardless of argument order, so it was not antisymmetric. Sorting over an inconsistent comparator is unspecified but safe, so this was latent rather than a live bug, but the new key removes the inconsistency by construction.

RyanJamesStewart:
Here is the harness that produced those numbers, in case it is useful to reproduce. It is a small example binary (not part of the PR) that reimplements both the old comparator and the new key so they can be A/B timed over the same fork set, hitting the real RocksDB reads each does. Run it with `cargo run --release --example bench_block_sort -- tests/data/sequential_blocks <height>` (e.g. height 105496 for the 7-block fork).

```rust
use std::cmp::Ordering;
use std::time::Instant;

use mina_indexer::{
    base::state_hash::StateHash,
    block::{
        precomputed::{PcbVersion, PrecomputedBlock},
        store::BlockStore,
        BlockComparison,
    },
    canonicity::{store::CanonicityStore, Canonicity},
    store::IndexerStore,
};
use tempfile::TempDir;

// Replica of the pre-PR free `block_cmp`: reads canonicity + comparison for
// BOTH args on every pairwise compare.
fn old_block_cmp(db: &IndexerStore, a: &StateHash, b: &StateHash) -> Ordering {
    let a_canonicity = db.get_block_canonicity(a).ok().flatten();
    let b_canonicity = db.get_block_canonicity(b).ok().flatten();
    let a_cmp = db.get_block_comparison(a).unwrap().unwrap();
    let b_cmp = db.get_block_comparison(b).unwrap().unwrap();

    match (a_canonicity, b_canonicity) {
        (Some(Canonicity::Canonical), _) => Ordering::Less,
        (_, Some(Canonicity::Canonical)) => Ordering::Greater,
        _ => a_cmp.cmp(&b_cmp),
    }
}

// Replica of the post-PR free `block_sort_key`: reads each input once.
fn new_block_sort_key(db: &IndexerStore, state_hash: &StateHash) -> (std::cmp::Reverse<bool>, BlockComparison) {
    let is_canonical = matches!(
        db.get_block_canonicity(state_hash).ok().flatten(),
        Some(Canonicity::Canonical)
    );
    let comparison = db.get_block_comparison(state_hash).unwrap().unwrap();
    (std::cmp::Reverse(is_canonical), comparison)
}

fn bench(label: &str, mut f: impl FnMut()) -> f64 {
    f(); // warm up
    let mut best = f64::MAX;
    for run in 0..3 {
        let t = Instant::now();
        for _ in 0..10_000 {
            f();
        }
        let per = t.elapsed().as_nanos() as f64 / 10_000.0 / 1000.0;
        eprintln!("  {label} run {run}: {per:.3} us/call");
        best = best.min(per);
    }
    eprintln!("{label} best: {best:.3} us/call");
    best
}

fn main() {
    let mut args = std::env::args().skip(1);
    let blocks_dir = args.next().expect("blocks dir");
    let height: u32 = args.next().expect("height").parse().unwrap();

    let temp = TempDir::new().unwrap();
    let store = IndexerStore::new(temp.path(), true).unwrap();

    // Ingest every block in the directory so block_comparison/canonicity CFs are populated.
    let mut at_height: Vec<StateHash> = Vec::new();
    let mut entries: Vec<_> = std::fs::read_dir(&blocks_dir)
        .unwrap()
        .filter_map(|e| e.ok().map(|e| e.path()))
        .filter(|p| p.extension().map(|x| x == "json").unwrap_or(false))
        .collect();
    entries.sort();
    for path in &entries {
        let pcb = PrecomputedBlock::parse_file(path, PcbVersion::V1).unwrap();
        let len = path.metadata().unwrap().len();
        store.add_block(&pcb, len).unwrap();
        if pcb.blockchain_length() == height {
            at_height.push(pcb.state_hash());
        }
    }
    assert!(
        at_height.len() > 1,
        "need a multi-block fork at height {height}, found {}",
        at_height.len()
    );
    eprintln!("fork at height {height}: {} blocks", at_height.len());

    let base = at_height.clone();
    let old = bench("OLD sort_by(block_cmp)        ", || {
        let mut v = base.clone();
        v.sort_by(|a, b| old_block_cmp(&store, a, b));
        std::hint::black_box(&v);
    });
    let new = bench("NEW sort_by_cached_key(key)   ", || {
        let mut v = base.clone();
        v.sort_by_cached_key(|sh| new_block_sort_key(&store, sh));
        std::hint::black_box(&v);
    });

    eprintln!("\nspeedup (old/new): {:.2}x", old / new);
}
```

